PR #97, opened 2020-05-08
Fix #96 added install_proprietary_server and codes_download_proprietary

### LINKED ISSUE #96: flashcap: install proprietary codes
Add support installing proprietary codes to install.sh. Clean up install_depot_server as well

robnagler:
Also added install_foss_server
install_depot_server takes an arg which means "force" to allow
install_foss_server to use
if $install_depot_server is set, it will override depot.radiasoft.org so
can test end-to-end.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -72,11 +72,12 @@ install_clean() {
 }
 
 install_depot_server() {
-    if [[ ${install_server:-github} != github ]]; then
+    local force=${1:-}
+    if [[ ! $force && ${install_server:-github} != github ]]; then
         echo -n "$install_server"
         return
     fi
-    echo -n https://depot.radiasoft.org
+    echo -n "$install_depot_server"
 }
 
 install_download() {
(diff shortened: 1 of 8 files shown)


bin/install.sh
```diff
@@ -72,11 +72,12 @@ install_clean() {
...
 
 install_depot_server() {
-    if [[ ${install_server:-github} != github ]]; then
+    local force=${1:-}
+    if [[ ! $force && ${install_server:-github} != github ]]; then
         echo -n "$install_server"
         return
     fi
-    echo -n https://depot.radiasoft.org
+    echo -n "$install_depot_server"
```
e-carlin, line 80: Clever. My bash has improved 2x in the past few days reading your code. Thanks.

bin/install.sh
```diff
@@ -161,6 +176,7 @@ install_init_vars() {
     : ${install_server:=github}
     : ${install_tmp_dir:=/var/tmp}
     : ${install_verbose:=}
+    : ${install_proprietary_key:=missing-proprietary-key}
```
e-carlin, line 179: Oof, Google doesn't like my searches for what this means. Thankfully [this](https://www.tldp.org/LDP/abs/html/special-chars.html) had the answer near the top of the page when just searching for ":".

PR #110, opened 2020-06-18
Fix radiasoft/sirepoflash#2. Update caplaser config.

e-carlin:
@robnagler the latest CapLaserBELLA-4.6.2.tar.gz is available. Should we automate this process so we don't have to pass files like this in the future? I built CapLaserBELLA-4.6.2.tar.gz from https://github.com/radiasoft/rsflash/tree/master/config/Bella

### DIFF
--- a/installers/rpm-code/codes/flash.sh
+++ b/installers/rpm-code/codes/flash.sh
@@ -43,10 +43,11 @@ flash_setup_CapLaserBELLA() {
     local type=$1
     codes_download_proprietary "flash/$type-4.6.2.tar.gz" "source"
     cd ..
-    ./setup "$type" -objdir="$type" -auto -2d -nxb=16 -nyb=16 +hdf5typeio \
+    ./setup "$type" -objdir="$type" -auto -2d -nxb=8 -nyb=8 +hdf5typeio \
             species=fill,wall +mtmmmt +usm3t +mgd mgd_meshgroups=6 \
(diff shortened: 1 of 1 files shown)


robnagler:
rscode-flash is built for alpha & dev.

I saw this when building CapLaserBELLA:

```txt
IGNORED SUGGESTIONS
Grid/GridParticles/GridParticlesMove suggested one of the following units as
well:
    IO/IOParticles
```

e-carlin:
@ncook882 does the above warning mean anything to you?

We are using the below setup command which is the same as the one as in [your bella.par](https://github.com/radiasoft/rsflash/blob/8f9a68b59c4187b0fdc6b13fb7ea817cc4ebeb3b/config/Bella/bella.par#L7).
```
./setup "$type" -objdir="$type" -auto -2d -nxb=8 -nyb=8 +hdf5typeio  species=fill,wall +mtmmmt +usm3t +mgd mgd_meshgroups=6  -parfile=bella.par +laser ed_maxPulses=1 ed_maxPulseSections=4  ed_maxBeams=1  -with-unit=physics/sourceTerms/Heatexchange/HeatexchangeMain/LeeMore
```

ncook882:
@e-carlin if the any of the various `Particles` units are used by the simulation, then it will want to have `IOParticles` included to handle writing them to file.

I don't think our simulations actually use this unit, albeit they may be included by default because we're not excluding them. Either way, this shouldn't have an issue on the simulation performance.

PR #115, opened 2020-08-05
WIP radiasoft/sirepo#2740. Use chrome-headless for testing

e-carlin:
@robnagler how do I test this to make sure it works? I've installed chrome myself and tested the sirepo side but I want to make sure these changes actually work.

### DIFF
--- a/installers/sirepo-dev/radiasoft-download.sh
+++ b/installers/sirepo-dev/radiasoft-download.sh
@@ -36,11 +36,11 @@ sirepo_dev_main() {
     done
     cd sirepo
     install_yum_install nodejs
-    for p in jshint karma karma-jasmine karma-phantomjs-launcher jasmine-core; do
-        if ! npm list "$p" >& /dev/null; then
-           npm install "$p" >& /dev/null
-        fi
-    done
+    npm install
+    install_tmp_dir
+    install_download https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm > chrome.rpm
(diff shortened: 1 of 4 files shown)


e-carlin:
Also, do I have to add the install of Chrome to a script for the build server? It will need to be installed there too.

installers/sirepo-dev/radiasoft-download.sh
```diff
@@ -36,11 +36,11 @@ sirepo_dev_main() {
...
     cd sirepo
     install_yum_install nodejs
-    for p in jshint karma karma-jasmine karma-phantomjs-launcher jasmine-core; do
-        if ! npm list "$p" >& /dev/null; then
-           npm install "$p" >& /dev/null
-        fi
-    done
+    npm install
+    install_tmp_dir
+    install_download https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm > chrome.rpm
```
robnagler, line 41: I think you can put this on one line, and then yum knows how to manage it:

```txt
install_yum_install https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm
```

You don't need install_tmp_dir then or even cd -

e-carlin: https://github.com/radiasoft/download/pull/115/commits/775fe9111be31abe3b18ecc1d47f558681eccd19 . I changed perl-dev too which is where I copied the code from.

PR #127, opened 2020-08-26
fix #126. Reload user for docker group permissions to take effect

### LINKED ISSUE #126: Need to log out and log in again after `radia_run redhat-docker`
```
sudo su - -c "radia_run redhat-docker"
<SNIP>
radiasoft$ docker run hello-world
docker: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/create: dial unix /var/run/docker.sock: connect: permission denied.
```
After logging out and in again I can use docker.

The docker install adds the user to the docker group. One must log out and log in again after being added to a group to gain the permissions.

e-carlin:
I think this is enough and will work with sudo - -c as well as running the installer as the root user. Not positive though.

### DIFF
--- a/installers/redhat-docker/radiasoft-download.sh
+++ b/installers/redhat-docker/radiasoft-download.sh
@@ -127,3 +127,4 @@ EOF2
     systemctl restart docker
     systemctl enable docker
 }
+su vagrant
--- a/installers/redhat-docker/radiasoft-download.sh
+++ b/installers/redhat-docker/radiasoft-download.sh
@@ -126,5 +126,5 @@ EOF2
     systemctl daemon-reload
     systemctl restart docker
     systemctl enable docker
-    su vagrant
+    install_msg 'You will need to logout completely, and log back in.'
 }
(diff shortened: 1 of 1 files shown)


installers/redhat-docker/radiasoft-download.sh
```diff
@@ -126,4 +126,5 @@ EOF2
     systemctl daemon-reload
     systemctl restart docker
     systemctl enable docker
+    su vagrant
```
robnagler, line 129: The command has to be run as root. Therefore, you have to stay as root. The shells would be nested:

1. User logs in as vagrant
2. User su's as root
3. This `su vagrant`

What this should be is:

```sh
install_msg 'You will need to logout completely, and log back in.'
```

e-carlin: https://github.com/radiasoft/download/pull/127/commits/ae4f9e0986d9d78a388c5fb1c27a3b252442e22e

PR #146, opened 2021-01-27
radiasoft/sirepoflash#16: flash rpm now contains just flash src

e-carlin:
@robnagler radiasoft/sirepo#3266 depends on this.

### DIFF
--- a/installers/rpm-code/codes/flash.sh
+++ b/installers/rpm-code/codes/flash.sh
@@ -1,30 +1,24 @@
 #!/bin/bash
 
 flash_main() {
+    local f="FLASH4.6.2"
     codes_dependencies hypre
-    codes_download_proprietary "flash/FLASH-4.6.2.tar.gz" "FLASH4.6.2"
+    codes_download_proprietary "flash/FLASH-4.6.2.tar.gz" "$f"
     flash_patch_makefile
-    install -d 755 "${codes_dir[share]}"/flash4
     local n
-    for n in CapLaser3D CapLaserBELLA RTFlame; do
-        flash_make_and_install_type "$n"
+    for n in CapLaser3D CapLaserBELLA; do
+        flash_download_simulation_source "$n"
     done
+    cd ..
+    mv "$f" "flash"
+    local a="flash.tar.gz"
+    tar --create  --gzip --file "$a" flash
+    install -d 755 "${codes_dir[share]}"/flash
+    install -m 444 "$a" "${codes_dir[share]}"/flash/
 }
 
 flash_download_simulation_source() {
     codes_download_proprietary "flash/$1-4.6.2.tar.gz" "source"
-    cd ..
-}
-
-flash_make_and_install_type() {
-    local type=$1
-    "flash_setup_$type" "$type"
-    cd "$type"
-    codes_make
-    # POSIT: sirepo.sim_data.flash.SimData.flash_exe_path assumes flash4-flashType
-    install -m 755 flash4 "${codes_dir[bin]}"/flash4-"$type"
-    # POSIT: sirepo.sim_data.flash.SimData.flash_setup_units_path assumes flash4/setup_units-flashType
-    install -m 444 setup_units "${codes_dir[share]}"/flash4/setup_units-"$type"
     cd ..
 }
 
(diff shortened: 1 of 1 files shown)


installers/rpm-code/codes/flash.sh
```diff
@@ -1,33 +1,27 @@
...
+    local f="FLASH4.6.2"
     codes_dependencies hypre
-    codes_download_proprietary "flash/FLASH-4.6.2.tar.gz" "FLASH4.6.2"
+    codes_download_proprietary "flash/FLASH-4.6.2.tar.gz" "$f"
     flash_patch_makefile
-    install -d 755 "${codes_dir[share]}"/flash4
     local n
-    for n in CapLaser3D CapLaserBELLA RTFlame; do
-        flash_make_and_install_type "$n"
+    for n in CapLaser3D CapLaserBELLA; do
```
robnagler, line 9: I thought we were getting rid of the types so these would be in the examples folder.

I think we have a problem with the examples in that we can release them in the sirepo image, because they contain modified proprietary code. correct?

installers/rpm-code/codes/flash.sh
```diff
@@ -1,33 +1,27 @@
...
+        flash_download_simulation_source "$n"
     done
+    cd ..
+    mv "$f" "flash"
+    local a="flash.tar.gz"
+    tar --create  --gzip --file "$a" flash
+    install -d 755 "${codes_dir[share]}"/flash
+    install -m 444 "$a" "${codes_dir[share]}"/flash/
```
robnagler, line 17: you don't need trailing slash

installers/rpm-code/codes/flash.sh
```diff
@@ -1,33 +1,27 @@
 #!/bin/bash
 
 flash_main() {
+    local f="FLASH4.6.2"
     codes_dependencies hypre
-    codes_download_proprietary "flash/FLASH-4.6.2.tar.gz" "FLASH4.6.2"
+    codes_download_proprietary "flash/FLASH-4.6.2.tar.gz" "$f"
```
robnagler, line 6: reuse $f in both places.

robnagler (changes requested):

installers/rpm-code/codes/flash.sh
e-carlin: That's coming in a future PR. radiasoft/sirepo#3266 still needs them.

Can you phrase your question another way? Are you asking if it is a problem if these examples are moved to lib files and thus become publicly available?

robnagler: yes, if they are lib files in the https://github.com/radiasoft/sirepo, they will be visible. We might be able to get the FLASH project to allow us to do this. @ncook882 what say you?

installers/rpm-code/codes/flash.sh
e-carlin: 0fde278

installers/rpm-code/codes/flash.sh
e-carlin: This required re-creating FLASH-4.6.2.tar.gz (dir contained within contained no `-` between FLASH and 4.6.2). You can find it in my homedir on you know what machine.

0fde278

installers/rpm-code/codes/flash.sh
e-carlin: It seems like a gray zone to me. The flash "kernel" won't be publicly available but the files that will be public are files that override certain flash functionality. I believe the only exposed things would be the names of functions in the flash kernel.

robnagler: The CapLaser files are copies of files in the FLASH distro with modifications. For example, source/Simulation/SimulationMain/CapLaser3D/MagneticResistivity_fullState.F90 is based off of source/physics/materialProperties/MagneticResistivity/MagneticResistivityMain/SpitzerHighZ/MagneticResistivity_fullState.F90.

We can't just distribute these as examples. Since you can't run flash without having the role, we'll have to distribute the examples in the rpm. They can get unpacked in the user's lib directory. When the role is dropped, maybe the entire user/db/xyzzy/flash directory has to disappear, because it contains the binary, lib files, etc.

ncook882: @robnagler it's not clear to me whether there's making available a single file would violate the user agreement, or if there is some threshold that would permit us making available the component files that differentiate individual configurations. I can follow up with Petros and others for their input on this matter.

Your suggestion of having the examples packaged with the source code seems consistent with our current authentication mechanism.

robnagler: @ncook882 Thanks for understanding. 

Just to clarify, we do not have rights to redistribute any part of the code, even a single line. This is not about copyright which has a threshold for fair use. The license is quite clear, in multiple places, including discussion of modifications.

robnagler (approved):
temporary merge due to path dep

PR #152, opened 2021-02-18
Fix radiasoft/sirepo#3353: $HOME was /root

e-carlin:
$HOME was /root but artifacts are in /home/vagrant.


robnagler (approved):

robnagler:
Should I rebuild flash-alpha.tar.gz?

e-carlin:
> Should I rebuild flash-alpha.tar.gz?

Yes, thanks.

PR #202, opened 2022-03-11
WIP radiasoft/container-jupyter-nvidia#9: compile hypre for gpu

e-carlin:
Same as rscode-elegant. Use the gpu-only flag to specify that gpu
specific compilation is needed.

Also, update hypre.

I tested the hypre change on opal and flash (the flash installer mentioned hypre). Both worked fine

### DIFF
--- a/installers/rpm-code/codes/hypre.sh
+++ b/installers/rpm-code/codes/hypre.sh
@@ -2,9 +2,25 @@
 
 hypre_main() {
     codes_dependencies common
-    codes_download https://github.com/hypre-space/hypre/archive/v2.11.2.tar.gz
+    codes_download https://github.com/hypre-space/hypre/archive/v2.24.0.tar.gz
     cd src
-    ./configure --prefix="${codes_dir[prefix]}"
+    local a=''
+    if [[ ${1:-} == 'gpu-only' ]]; then
+        #POSIT: container-jupyter-nvidia (cuda-11.1)
+        # and running on Tesla V100 (arch 70)
+        a=(
+            --with-mpi
+            --enable-unified-memory
+            --prefix="${codes_dir[prefix]}"
+            --with-MPI
+            --with-cuda
+            --with-cuda-home='/usr/local/cuda-11.1'
+            --with-gpu-arch='70'
+        )
+    fi
+    ./configure \
+        --prefix="${codes_dir[prefix]}" \
+        "${a[@]}"
     # hypre install does a chmod -R on install dirs, which
     # makes a mess of things so install manually.
     codes_make all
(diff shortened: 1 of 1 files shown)


installers/rpm-code/codes/hypre.sh
```diff
@@ -2,9 +2,25 @@
 
 hypre_main() {
     codes_dependencies common
-    codes_download https://github.com/hypre-space/hypre/archive/v2.11.2.tar.gz
+    codes_download https://github.com/hypre-space/hypre/archive/v2.24.0.tar.gz
     cd src
-    ./configure --prefix="${codes_dir[prefix]}"
+    local a=''
```
robnagler, line 7: `local -a a=()`

installers/rpm-code/codes/hypre.sh
```diff
@@ -2,9 +2,25 @@
...
+    if [[ ${1:-} == 'gpu-only' ]]; then
+        #POSIT: container-jupyter-nvidia (cuda-11.1)
+        # and running on Tesla V100 (arch 70)
+        a=(
+            --with-mpi
+            --enable-unified-memory
+            --prefix="${codes_dir[prefix]}"
```
robnagler, line 14: duplicate? it's passed below.

installers/rpm-code/codes/hypre.sh
```diff
@@ -2,9 +2,25 @@
...
-    codes_download https://github.com/hypre-space/hypre/archive/v2.11.2.tar.gz
+    codes_download https://github.com/hypre-space/hypre/archive/v2.24.0.tar.gz
     cd src
-    ./configure --prefix="${codes_dir[prefix]}"
+    local a=''
+    if [[ ${1:-} == 'gpu-only' ]]; then
+        #POSIT: container-jupyter-nvidia (cuda-11.1)
+        # and running on Tesla V100 (arch 70)
+        a=(
+            --with-mpi
```
robnagler, line 12: Is hypre compiled without mpi normally? Seems like we would want mpi.

installers/rpm-code/codes/hypre.sh
```diff
@@ -2,9 +2,25 @@
...
+            --enable-unified-memory
+            --prefix="${codes_dir[prefix]}"
+            --with-MPI
```
robnagler, line 15: is this a duplicate of with-mpi?

installers/rpm-code/codes/hypre.sh
```diff
@@ -2,9 +2,25 @@
...
+            --with-cuda
+            --with-cuda-home='/usr/local/cuda-11.1'
```
robnagler, line 17: You don't need quotes. Would be nice to get this from somewhere, but I know that's hard.

robnagler (changes requested):

installers/rpm-code/codes/hypre.sh
e-carlin: 495c144

installers/rpm-code/codes/hypre.sh
e-carlin: 495c144

robnagler (approved):

PR #204, opened 2022-03-29
Fix#201: build openmc

e-carlin:
In addition, modify codes_cmake_fix_lib_dir to account for submodules
with their own makefiles. For example, openmc has [submodules](https://github.com/openmc-dev/openmc/tree/2f068863d032d56152505c7d8ff6b969f4705003/vendor) 
and they have [their own CMakeLists.txt](https://github.com/fmtlib/fmt/blob/d141cdbeb0fb422a3fb7173b285fd38e0d1772dc/CMakeLists.txt)

### DIFF
--- a/installers/rpm-code/codes.sh
+++ b/installers/rpm-code/codes.sh
@@ -27,7 +27,7 @@ codes_cmake_fix_lib_dir() {
     perl -pi -e '/include\(GNUInstallDirs/ && ($_ .= q{
 set(CMAKE_INSTALL_LIBDIR "lib" CACHE PATH "Library installation directory." FORCE)
 GNUInstallDirs_get_absolute_install_dir(CMAKE_INSTALL_FULL_LIBDIR CMAKE_INSTALL_LIBDIR)
-})' CMakeLists.txt
+})' $(find . -name CMakeLists.txt)
 }
 
 codes_curl() {
@@ -102,6 +102,7 @@ codes_download() {
                 git clone $r -q "$repo"
                 cd "$d"
                 git checkout "$qualifier"
+                git submodule update --init --recursive
             else
                 git clone $r --depth 1 "$repo"
                 cd "$d"
--- a/installers/rpm-code/codes.sh
+++ b/installers/rpm-code/codes.sh
@@ -24,10 +24,10 @@ codes_cmake() {
 
 codes_cmake_fix_lib_dir() {
     # otherwise uses ~/.local/lib64
-    perl -pi -e '/include\(GNUInstallDirs/ && ($_ .= q{
+    find . -name CMakeLists.txt -print0 | xargs -0 perl -pi -e '/include\(GNUInstallDirs/ && ($_ .= q{
 set(CMAKE_INSTALL_LIBDIR "lib" CACHE PATH "Library installation directory." FORCE)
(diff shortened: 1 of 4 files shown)


installers/rpm-code/codes.sh
```diff
@@ -27,7 +27,7 @@ codes_cmake_fix_lib_dir() {
     perl -pi -e '/include\(GNUInstallDirs/ && ($_ .= q{
 set(CMAKE_INSTALL_LIBDIR "lib" CACHE PATH "Library installation directory." FORCE)
 GNUInstallDirs_get_absolute_install_dir(CMAKE_INSTALL_FULL_LIBDIR CMAKE_INSTALL_LIBDIR)
-})' CMakeLists.txt
+})' $(find . -name CMakeLists.txt)
```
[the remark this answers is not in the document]
robnagler: If there are spaces in directories, this will break. It needs to be done with xargs, e.g.

```sh
find ... -print0 | xargs -0 perl -pi ... 
```

I don't think this will break anything. pymesh(?) is the only one with lots of makefiles. 

This is where we need a build test system, because we really should rebuild all codes with this fixup to verify they build correctly. Only codes with more than one CMakeLists.txt matter so that's a check we could add in this script and output a warning if the files actually differ. 

You could change the `-pi` to `.pi.codes_fix`. Then do something like this after the edit:

```sh
find ... -print0 | xargs -0 -n 1 -I % diff -c %.codes_fix %
```

If there is any output, you'll at least see if it is more than one file, and if it is screwed up.

installers/rpm-code/codes.sh
```diff
@@ -102,6 +102,7 @@ codes_download() {
                 git clone $r -q "$repo"
                 cd "$d"
                 git checkout "$qualifier"
+                git submodule update --init --recursive
```
robnagler, line 105: What are the submodules. I see you are building quite a few other codes.

installers/rpm-code/codes/openmc.sh
```diff
@@ -0,0 +1,48 @@
...
+        -D CMAKE_INSTALL_PREFIX="${codes_dir[prefix]}" \
+        -D HDF5_INCLUDE_DIRS="${codes_dir[prefix]}" \
+        -D HDF5_PREFER_PARALLEL=on \
+        -D OPENMC_USE_DAGMC=on
+    codes_make
+    codes_make_install
+}
+
+openmc_python_install() {
+    install_pip_install git+https://github.com/openmc-dev/openmc.git@develop
```
robnagler, line 47: Possibly share "develop" in a global constant since if they get out of sync, there could be a problem.

installers/rpm-code/dev-build-all.sh
```diff
@@ -44,6 +44,7 @@ codes=(
 # see pyopaltools.sh
 #    pyopaltools
     opal
+    openmc
```
robnagler, line 47: Note: also update the "all" on the build machine, too, when this is released to alpha.

robnagler (changes requested):

installers/rpm-code/codes.sh
e-carlin: https://github.com/openmc-dev/openmc/tree/2f068863d032d56152505c7d8ff6b969f4705003/vendor

moellep:
@robnagler is this ready to be approved? Stephen asked about this for CloudMC

PR #226, opened 2022-06-06
fix #225: ci-pull-request supports test.sh

### LINKED ISSUE #225: ci-pull-request supports test.sh
If there is a test.sh, run that instead of pykern test.

rorour:


installers/ci-pull-request/radiasoft-download.sh
```diff
@@ -44,7 +44,7 @@ ci_pull_request_main() {
             fi
             pip uninstall -y '$r' >& /dev/null || true
             pip install -e .
-            pykern test
+            [ -f test.sh ] && bash test.sh || pykern test
```
robnagler, line 47: The nice thing about shell programming is that there are so many ways to do this... We have moved to a model where we spell things out with if/then/else. With Bash, we use [double square brackets](http://mywiki.wooledge.org/BashFAQ/031) which are more robust and modern:

```bash
if [[ -f test.sh ]]; then
    bash test.sh
else
    pykern test
fi
```

robnagler (changes requested):
Correct but we have a different style.

robnagler (approved):

PR #238, opened 2022-06-21
fix #237: dots are not escaped in regex match

### LINKED ISSUE #237: Dots are not escaped in regex match
https://github.com/radiasoft/download/blob/84922815033e849b0f017102cb4e30572d98bd67/bin/install.sh#L83

This regex match will unintentionally match domains such as `apixgithub.com`.

rorour:


bin/install.sh
```diff
@@ -80,7 +80,7 @@ install_download() {
     fi
     local x=( "${install_curl_flags[@]}" )
     install_info curl ${x[@]+"${x[@]}"} "$url"
-    if [[ $url =~ https://api.github.com ]]; then
+    if [[ $url =~ https://api\.github\.com ]]; then
```
robnagler, line 83: Add `^https`, too.

robnagler (changes requested):

robnagler (approved):
Nice find, btw.

PR #239, opened 2022-06-21
fix #236: sirepo-dev install package.json globally

### LINKED ISSUE #236: sirepo-dev install package.json globally
https://github.com/radiasoft/sirepo/pull/4536#discussion_r899635702 

Add the global install similar to what sirepo-ci does but it needs to be idempotent (similar to what test.sh does):

```sh
if ! [[ $(type -p karma) && $(type -p jshint) ]]; then
     cd ~/src/radiasoft/sirepo
     eval npm install -g $(jq -M -r '.devDependencies | keys | @sh' package.json)
fi
```

Possibly create a new function in sirepo-dev installer so modular.

rorour:


installers/sirepo-dev/radiasoft-download.sh
```diff
@@ -40,4 +40,13 @@ sirepo_dev_main() {
     cd sirepo
     npm install
     cd ..
+    sirepo_dev_npm
+}
+
+sirepo_dev_npm() {
+    if ! [[ $(type -p karma) && $(type -p jshint) ]]; then
+       cd ~/src/radiasoft/sirepo
```
robnagler, line 48: Can be removed.

installers/sirepo-dev/radiasoft-download.sh
```diff
@@ -40,4 +40,13 @@ sirepo_dev_main() {
     cd sirepo
     npm install
     cd ..
+    sirepo_dev_npm
```
robnagler, line 43: Move one line up so you don't need to cd below.

robnagler (changes requested):

installers/sirepo-dev/radiasoft-download.sh
rorour: 09ed5adf527e153ce4b1c312f35d0f1542891da6

installers/sirepo-dev/radiasoft-download.sh
rorour: 09ed5adf527e153ce4b1c312f35d0f1542891da6

robnagler (approved):

PR #257, opened 2022-06-24
Fix #233 installer for mantid

### LINKED ISSUE #233: mantid installer
https://developer.mantidproject.org/BuildingWithCMake.html#buildingwithcmake

e-carlin:


installers/rpm-code/codes/mantid.sh
```diff
@@ -0,0 +1,93 @@
...
+    codes_make
+    codes_make_install
+    cd ../
+}
+
+mantid_patch_eigen_cmake() {
+    # TODO(e-carlin): f32 eigen3 is 3.3.  Mantid technically depends on Eigen3 3.4 but the change
+    # that bumped 3.3 to 3.4 (6d5aa8a43d980d2ea0cf919da64fd3a2c44def0d) made no other other changes
+    # to the code. So, use 3.3 for now and see if there are any issues.
+    echo 'find_package(Eigen3 3.3 REQUIRED)' > ./buildconfig/CMake/Eigen.cmake
```
e-carlin, line 92: @robnagler will this work at all or do you foresee problems?

robnagler: There will always be problems, but you can't predict them. If this works reliably for now, then try this.

installers/rpm-code/codes/mantid/rsmantid.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+set -euo pipefail
+exec ${MANTID_PYTHON_SCRIPT_INSTALL_LOCATION} --classic
```
robnagler, line 3: Always quote unless it's necessary not to. Everything is an example. Use POSIT or ASSUME if you are not quoting something.

robnagler (changes requested):

installers/rpm-code/codes/ipykernel.sh
```diff
@@ -3,25 +3,28 @@
...
     codes_dependencies common
     install_pip_install ipykernel
-    local i=3
-    local v=py$i
     # http://ipython.readthedocs.io/en/stable/install/kernel_install.html
     # http://www.alfredo.motta.name/create-isolated-jupyter-ipython-kernels-with-pyenv-and-virtualenv/
-    local where=( $(python -m ipykernel install --display-name "Python $i" --name "$v" --user) )
+    local where=( $(ipykernel_install) )
     where=${where[-1]}
     if [[ ! $where  =~ ${codes_dir[share]} ]]; then
```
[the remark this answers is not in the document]
e-carlin: I don't see how it is unused? It is used here on line 10 and then again on line 11.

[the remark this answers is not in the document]
e-carlin: With the other changes it made sense to move the code around 51ba297

installers/rpm-code/codes/ipykernel.sh
```diff
@@ -3,25 +3,28 @@
...
-                    qw(LD_LIBRARY_PATH PKG_CONFIG_PATH PYENV_VERSION PYTHONPATH),
-                ),
-            );
-        }
-        s/^\{/{\n "env": {\n@{[_e()]}\n },/
-    ' "${where}"/kernel.json
+}
+
+ipykernel_install() {
+    local i=3
```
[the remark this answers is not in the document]
e-carlin: 51ba297

installers/rpm-code/codes/mantid.sh
e-carlin: Thanks

installers/rpm-code/codes/mantid.sh
```diff
@@ -0,0 +1,93 @@
...
+    codes_cmake_build
+    codes_make_install
+}
+
+mantid_install_rsmantid() {
+    local s=rsmantid
+    codes_download_module_file "$s.sh"
+    local l="$1/bin/mantidpython"
+    MANTID_PYTHON_SCRIPT_INSTALL_LOCATION="$l" \
+    perl -p -e 's/\$\{(\w+)\}/$ENV{$1} || die("$1: not found")/eg' "$s.sh" \
```
robnagler, line 45: I indent this line at 4 and the next line leave alone.

installers/rpm-code/codes/mantid.sh
```diff
@@ -0,0 +1,93 @@
...
+    local install_prefix="$1"
+    codes_download mantidproject/mantid
+    mantid_patch_eigen_cmake
+    codes_cmake \
+        -DBOOST_ROOT="${codes_dir[prefix]}" \
+        -DCMAKE_INSTALL_PREFIX="$install_prefix" \
+        -DENABLE_WORKBENCH=OFF \
+        -DMANTID_FRAMEWORK_LIB=BUILD \
+        -DMANTID_QT_LIB=OFF \
+        -DNEXUS_CPP_LIBRARIES="${codes_dir[lib]}/libNeXusCPP.so" \
```
robnagler, line 32: This is always interesting. The only thing needing quoting is between the braces. Some people there is some special syntax with `--key=value` args, but the shell knows nothing about this (caveat for interactive autocomplete) so I would say as an "example" it's better to quote just what's needed when there's only one thing quoted. Obviously with multiple substitutions, then quote the whole arg (including --key).

installers/rpm-code/codes/mantid.sh
```diff
@@ -0,0 +1,93 @@
...
+        -G'Unix Makefiles'
+    codes_cmake_build
+    codes_make_install
+}
+
+mantid_install_rsmantid() {
+    local s=rsmantid
+    codes_download_module_file "$s.sh"
+    local l="$1/bin/mantidpython"
+    MANTID_PYTHON_SCRIPT_INSTALL_LOCATION="$l" \
```
robnagler, line 44: a case could be made to move build_replace_vars from containers/bin/build.sh (leave an alias) to download/bin/install.sh. Maybe another time.

installers/rpm-code/codes/mantid.sh
```diff
@@ -0,0 +1,93 @@
...
+        -DNEXUS_INCLUDE_DIR="${codes_dir[include]}/nexus/" \
+        -G'Unix Makefiles'
+    codes_cmake_build
+    codes_make_install
+}
+
+mantid_install_rsmantid() {
+    local s=rsmantid
+    codes_download_module_file "$s.sh"
+    local l="$1/bin/mantidpython"
```
robnagler, line 43: I wouldn't use a temporary for this. It doesn't improve readability. Just echo $MANTID_PYTHON_SCRIPT_INSTALL_LOCATION (below)

installers/rpm-code/codes/mantid.sh
```diff
@@ -0,0 +1,93 @@
...
+
+-LD_PRELOAD=${LOCAL_PRELOAD} \
++echo LD_PRELOAD=${LOCAL_PRELOAD} \
+     PYTHONPATH=${LOCAL_PYTHONPATH} \
+     QT_API=${LOCAL_QT_API} \
+EOF
+    local -a e=( $(bash "$p"))
+    e=( "${e[@]::${#e[@]}-5}" )
+    local -a r=()
+    for v in "${e[@]}"; do
```
robnagler, line 65: local v

installers/rpm-code/codes/mantid.sh
```diff
@@ -0,0 +1,93 @@
...
+        -DNEXUS_CPP_LIBRARIES="${codes_dir[lib]}/libNeXusCPP.so" \
+        -DNEXUS_C_LIBRARIES="${codes_dir[lib]}/libNeXus.so" \
+        -DNEXUS_INCLUDE_DIR="${codes_dir[include]}/nexus/" \
+        -G'Unix Makefiles'
+    codes_cmake_build
+    codes_make_install
+}
+
+mantid_install_rsmantid() {
+    local s=rsmantid
```
robnagler, line 41: I try to always declare args:

```sh
local install_d=$1
```

installers/rpm-code/codes/mantid.sh
```diff
@@ -0,0 +1,93 @@
...
+    # Needs to be in same dir as $1 because script uses relative paths
+    local p="$1-$RANDOM"
+    patch --quiet --output="$p" "$1" <<'EOF'
+@@ -54,5 +54,5 @@
+ fi
+
+-LD_PRELOAD=${LOCAL_PRELOAD} \
++echo LD_PRELOAD=${LOCAL_PRELOAD} \
+     PYTHONPATH=${LOCAL_PYTHONPATH} \
+     QT_API=${LOCAL_QT_API} \
```
robnagler, line 60: Is QT_API needed?

installers/rpm-code/codes/mantid.sh
```diff
@@ -0,0 +1,93 @@
...
+EOF
+    local -a e=( $(bash "$p"))
+    e=( "${e[@]::${#e[@]}-5}" )
```
robnagler, line 63: You don't need to rewrite e. Just put it in place. Less code. 

However, this is where a little copying than a lot of code. There's a lot of complexity in what's going on. Do we know the basic relative dirs for the paths and can set PYTHONPATH LD_PRELOAD (and LD_LIBRARY_PATH?) manually without all these evals?

robnagler (changes requested):

installers/rpm-code/codes/mantid.sh
e-carlin: Can you show me what you mean? `$MANTID_PYTHON_SCRIPT_INSTALL_LOCATION` is just set on the perl command. Should I define it as a local variable instead of "$l"?

installers/rpm-code/codes/mantid.sh
e-carlin: Can you give me the syntax for that? `for v in ( ${e[@]::${#e[@]}-5} ); do` doesn't work.

What do you mean by `basic relative dirs`? I can find out what PYTHONPATH etc end up being set to and set them myself but I preferred this way because:
1. If they change the script to add new values maybe we'll pick them up
2. If they change the script maybe the patch will break and we'll be tipped off that something has changed
3. It brings the readers attention to the mantidpython script and that it holds information about what environment is needed to get mantid running. So, it is a breadcrumb to help with debugging in the future. 

Obviously none of that is foolproof but I think it is better than the flipside which is we'll recompile mantid in the future, it will build fine but someone will tell us it isn't working right, we'll have to trace it back down to some environment issue and I'll have to ask myself how I determined this environment in the first place.

installers/rpm-code/codes/mantid.sh
e-carlin: Probably not (we don't build with qt enabled) but what is gained by removing it?

installers/rpm-code/codes/mantid/rsmantid.sh
e-carlin: 8c3e1a0

installers/rpm-code/codes/mantid.sh
e-carlin: 8c3e1a0

installers/rpm-code/codes/mantid.sh
e-carlin: 8c3e1a0 thanks for the details. I wasn't sure what our preferred style was...

installers/rpm-code/codes/mantid.sh
e-carlin: 8c3e1a0

installers/rpm-code/codes/mantid.sh
e-carlin: 8c3e1a0

e-carlin:
@robnagler all comments addressed. Ready for another review.

e-carlin:
@robnagler if it is ok with you in the future will you wait until I tell you that I'm ready for another review before reviewing. It makes it hard for me to keep track of things when I'm working on 2 reviews at once.

installers/rpm-code/codes/mantid.sh
robnagler: Can you attach a copy of mantidpython after it has been generated to here? I'd like to see if what's going on is what I think's going on. :)

robnagler:
Will do. I wasn't clear, and sometimes people forget to press "request review" on re-reviews.

installers/rpm-code/codes/mantid.sh
e-carlin: [Here](https://github.com/radiasoft/download/files/8993130/mantidpython.txt) it is. FWIW, [this is the template](https://github.com/radiasoft/download/files/8993140/mantidpython.in.txt) that file is generated from.

robnagler: `for v in ( ${e[@]::${#e[@]}-5} ); do` is too much try `for v in "${e[@]::${#e[@]}-5}"`.  Parens for arrays are only meaningful in some contexts.

e-carlin:
@robnagler not sure where you are on this but I believe I've addressed all of your comments.

robnagler (commented):
Made several changes in 1fd4074.  The context of the $LOCAL eval is build time not runtime. This is important. We can discuss. Please review/test my changes.

e-carlin:
Thanks, I left some comments on that commit. Building the rpm to test now...

e-carlin:
@robnagler ready for another review

installers/rpm-code/codes/ipykernel.sh
```diff
@@ -3,25 +3,25 @@
 ipykernel_main() {
     codes_dependencies common
     install_pip_install ipykernel
-    local i=3
-    local v=py$i
+}
+
+ipykernel_python_install() {
+    local -a e=()
+    for x in LD_LIBRARY_PATH PKG_CONFIG_PATH PYTHONPATH; do
```
robnagler, line 10: I don't think these values can come from the build environment. Maybe PKG_CONFIG_PATH. They normally are inherited via ~/.bashrc so we don't need to set them explicitly. Has something changed?

robnagler (changes requested):

installers/rpm-code/codes/ipykernel.sh
e-carlin: Thanks, I see now how these come from home-env.

Does the jupyterhub process have the correct env? I think we need to use https://jupyterhub.readthedocs.io/en/stable/api/spawner.html#jupyterhub.spawner.Spawner.env_keep because [this is](https://github.com/jupyterhub/jupyterhub/blob/909b3ad4d708a7369343c8a8537c95c05aae3b67/jupyterhub/spawner.py#L433) only env that is passed to the notebook which doesn't contain `LD_LIBRARY_PATH`

e-carlin: aaaeb0f

installers/rpm-code/codes/mantid.sh
```diff
@@ -0,0 +1,86 @@
...
+mantid_ipykernel() {
+    local INSTALLDIR=$1
+    local mantidpython=$2
+    local LOCAL_PRELOAD LOCAL_PYTHONPATH LOCAL_LDPATH
+    eval "$(grep -P '^LOCAL_\w+=' "$mantidpython")"
+    python -m ipykernel install \
+        --display-name 'Mantid Python' \
+        --name pymantid \
+        --user \
+        --env PYENV_VERSION py3 \
```
robnagler, line 62: I don't think PYENV_VERSION is required, because it's the default. We can avoid distributing the coupling with `py3` by not specifying it here.

robnagler (changes requested):

installers/rpm-code/codes/mantid.sh
robnagler: > Probably not (we don't build with qt enabled) but what is gained by removing it?

Less is more. The more configuration, the more complex, which makes it harder to read and understand, and introduces unnecessary implicit coupling.

installers/rpm-code/codes/mantid.sh
e-carlin: 39360a2

robnagler (approved):

PR #263, opened 2022-06-30
Fix #261: add installer for MGIS

### LINKED ISSUE #261: MGIS installer
@nickgoldring writes:

> I would like to install MFrontGenericInterfaceSupport on RS Jupyter server in order to extend the thermal simulations to include mechanical analysis. Installation instructions can be found here: https://thelfer.github.io/mgis/web/install.html.
> robnagler , please let me know if this is something that I can help to get installed. In particular, I want to use the mgis.fenics module described here: https://thelfer.github.io/mgis/web/mgis_fenics.html.

e-carlin:


installers/rpm-code/codes/mgis.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+mgis_main() {
+    codes_dependencies common boost fenics
+    codes_download thelfer/MFrontGenericInterfaceSupport
+    local f
+    for f in $(find . -name 'CMakeLists.txt'); do
```
e-carlin, line 7: @robnagler thanks for your comment regarding us having support for c++17. This is the simplest hack I could find to get it working. `set(CMAKE_CXX_STANDARD 17)\nset(CXX_STANDARD_REQUIRED ON)` are in the top-level CMakeList.txt. But for an uknown (to me) reason that setting doesn't propagate down to the children CMakeLists.txt. This forces that to happen. I'm curious if there is a better way to do this?

robnagler: I do not know CMake. If you made it work, more power to you. :)

robnagler (approved):

PR #271, opened 2022-07-12
fix #270 rpm-code/dev-env.sh look for ~/src so relocatable

### LINKED ISSUE #270: make rpm-code/dev-env.sh relocatable
Instead of hardwiring $HOME it should verify the path contains src/radiasoft/download/installers/rpm-code.

robnagler:


installers/rpm-code/dev-env.sh
```diff
@@ -1,12 +1,27 @@
 #!/bin/bash
 # intended to be sourced
+cd "$(dirname "${BASH_SOURCE[0]}")"
+_root() {
+    local r=$PWD
+    while [[ $(basename $r) != src ]]; do
+        d=$(dirname "$r")
+        echo $d
```
rorour, line 8: I just tested this and it seemed like all these extra echos are being added to the value of repo_fedora_dir

installers/rpm-code/dev-env.sh
```diff
@@ -1,12 +1,27 @@
...
+    done
+    echo $r
+}
 export dev_port=2916
 export install_server=http://$(hostname -f):$dev_port
 export fedora_version=32
-export repo_rel_dir=fedora/$fedora_version/$(arch)/dev
-export rpm_code_install_dir=$HOME/src/yum/$repo_rel_dir
-export radiasoft_repo_file=$(dirname $(dirname $rpm_code_install_dir))/radiasoft.repo
+export repo_fedora_dir=$(_root)/src/yum/fedora
```
rorour, line 21: I think this src/ isn't needed

installers/rpm-code/dev-env.sh
robnagler: 5fedd36

installers/rpm-code/dev-env.sh
robnagler: 5fedd36

Also fixed an error in rpm_perl_install_dir.

Thanks for testing and reviewing!

rorour (approved):
Looks good

PR #314, opened 2022-07-29
fix #313: vagrant ssh expects cmds with -c or STDIN

### LINKED ISSUE #313: vagrant ssh error
I ran `vagrant-rsconf-dev` on centos. My guess is this is due to the change to direct `vagrant ssh`. 
```
+ vagrant ssh sudo su -
The machine with the name 'sudo' was not found configured for
this Vagrant environment.
+ install_err_trap
```

e-carlin:


robnagler (commented):
That's not going to work. You need vagrant ssh -c 'sudo su -'

I have that in 229-docker-disk and other fixes which i'm testing.

e-carlin:
Ok, closing and I'll let you fix there.

Can you elaborate on why `-c` is needed instead of adding to heredoc? I ran it and it seemed to work fine.

robnagler:
> Can you elaborate on why -c is needed instead of adding to heredoc? I ran it and it seemed to work fine.

Because this PR didn't add a heredoc, just the command. Two heredocs would be fine, but -c is simpler in this case.

e-carlin:
> Two heredocs would be fine, but -c is simpler in this case.

I'm sorry but can you explain a little more? My command "works" (as far as I can tell) so I'm having trouble identifying what is wrong with it.

What is the difference between adding to the heredoc as I did vs having two heredocs or -c? My guess is that it has to do with the context in which things are run in but I don't know how to test that theory. 

Here is what I've tried: 
```
# Testing if there is difference in the envs
v$ vagrant ssh -c 'sudo su -' > c-env <<EOF
env
EOF
v$ vagrant ssh > e-env <<EOF
> sudo su -
> env
> EOF
v$ diff c-env e-env
1c1
< Last login: Fri Jul 29 20:16:24 UTC 2022
---
> Last login: Fri Jul 29 20:16:35 UTC 2022
# None found
# Testing login vs non-login shell
v$ vagrant ssh -c 'sudo su -' <<EOF
> echo $0
> EOF
Last login: Fri Jul 29 20:17:14 UTC 2022
-bash
v$ vagrant ssh <<EOF
> sudo su -
> echo $0
> EOF
Last login: Fri Jul 29 20:19:45 UTC 2022
-bash
# No difference
# I also thought maybe it has something to do with password and prompting but passwords are disabled for sudo 
```

robnagler:
Egads! Of course it works! Your solution is fine. I have never seen that particular construct: a little bit like tail recursion. I will keep it in mind in other places. Thanks for the lesson.

e-carlin:
Aha, thanks:)

PR #316, opened 2022-07-29
Fix #299 remove docker disk creation

### LINKED ISSUE #224: vagrant.vbguest.no_install = true
```ruby
    if Vagrant.has_plugin? "vagrant-vbguest"
      config.vbguest.no_install  = true
      config.vbguest.auto_update = false
      config.vbguest.no_remote   = true
    end
```

### LINKED ISSUE #299: Remove docker disk from redhat-docker installer
Perhaps make backward compatible. Unless you are building beamsim-jupyter, perhaps. I think we could increase the size of the virtualbox root disk.

### LINKED ISSUE #315: vagrant sometimes inserts: 127.0.1.1 v6.radia.run v6
This one is a strange one, which we handle, but not generally enough. Note the IP address is 127.0.**1**.1. Need to delete this line as well.

robnagler:
Fix #224 default to no vbguest
Fix #315 relax /etc/hosts pattern

### DIFF
--- a/installers/redhat-docker/radiasoft-download.sh
+++ b/installers/redhat-docker/radiasoft-download.sh
@@ -24,7 +24,6 @@ rm -rf /var/lib/docker/*
 umount /var/lib/docker
 rmdir /var/lib/docker
 perl -pi -e "s{^/var/lib/docker.*}{}" /etc/fstab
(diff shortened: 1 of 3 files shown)


robnagler:
needed to merge

PR #318, opened 2022-08-01
Fix #317 radiasoft.repo comes from version

### LINKED ISSUE #317: repo file needs to be in version id

robnagler:


robnagler:
emergency fix

e-carlin:
lgtm. A search across our code for `repo_fedora_dir` makes me believe that var is only used in that file. Does it need to be exported?

robnagler:
Thanks. Fix 5560f30 on fedora branch

PR #324, opened 2022-08-06
Fix #321 lib0 node14

### LINKED ISSUE #321: lib0 no longer supports nodejs 12
In fedora32 the default nodejs is v12.22.1. But lib0 [dropped support for it](https://github.com/dmonad/lib0/commit/deea8b173395b216af8f52482d7c063a7600847f). lib0 is a nested dependency of jupyter_rs_vtk (at least, probably all jupyter widgets depend on it). 

I think we can specify nodejs:16/default

e-carlin:
lib0 expects >= node:14 so use node:16. Other fixes encountered when running dev-build-all.sh

### DIFF
--- a/installers/beamsim-codes/radiasoft-download.sh
+++ b/installers/beamsim-codes/radiasoft-download.sh
@@ -45,7 +45,6 @@ declare -a _beamsim_codes_all=(
 
     madx
     mantid
-    mgis
     openmc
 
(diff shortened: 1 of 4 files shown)


e-carlin:
Emeregency fix so merged but please review.

robnagler (commented):
lgtm

PR #325, opened 2022-08-06
The module command is not available on all systems. Those with dnf are new enough to have it

e-carlin:


e-carlin:
Emergency merge.

installers/redhat-base/radiasoft-download.sh
```diff
@@ -110,5 +110,7 @@ EOF
     install_yum_install "${x[@]}"
     # See: git.radiasoft.org/download/issues/231
     # Remove when we upgrade to F36 (has a newer node)
-    install_sudo yum module enable -y nodejs:16/default
+    if [[ $(type -t dnf) ]]; then
```
robnagler, line 113: Sorry, I should have caught this. 

```sh
if [[ $install_os_release_id == fedora && $install_os_release_version_id == 32 ]]; then
    install_sudo dnf module enable -y nodejs:16/default
fi
```

e-carlin: No worries, https://github.com/radiasoft/download/pull/327

PR #333, opened 2022-08-25
Fix #332 force h5py to 2.10.0

### LINKED ISSUE #332: tensorflow requires older h5py
Our tensorflow doesn't work with the version of h5py we have installed:

When you pip install h5py, there is this error.
tensorflow 2.3.1 requires h5py<2.11.0,>=2.10.0, but you have h5py 3.7.0 which is incompatible.

I think we need to fix the h5py version, or upgrade tensorflow.

robnagler:


robnagler:
@moellep @gurhar1133 you should be able to dnf update rscode-common to get h5py==2.10.0

PR #372, opened 2022-11-07
fix #371: install_repo_eval replaces install_repo everywhere

### LINKED ISSUE #371: 20221130 install_repo_eval replaces install_repo everywhere
I thought install_repo ran as a subshell, but it doesn't so you are right about install_repo_eval replacing install_repo everywhere. We should rename install_repo to something install_repo_internal so that we don't use it anyway. 

The original idea was that install_repo_eval would run in the current shell.  It was added in 569c8238d36d729e6edd244cdaaf54420d22216f, and it seems like it should always be used for install_repo_eval. I think there was some code at the time that required you to call install_repo by setting up variables and then running in a subshell. 

We could rename it. I'm not attached.

_Originally posted by @robnagler in https://github.com/radiasoft/download/issues/272#issuecomment-1295593259_

rorour:
@robnagler I don't have permission to push to [radiasoft/container-python3](https://github.com/radiasoft/container-python3), but [install_repo should be replaced with install_repo_eval](https://github.com/radiasoft/container-python3/blob/20bd470e4a3d8eb6d08a0e60c9cf7143f52ebea7/container-conf/build.sh#L7) before this is merged.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -248,41 +248,6 @@ install_os_release_vars() {
     fi
 }
 
-install_repo() {
-    if (( $# > 0 )); then
-        install_repo=$1
-        shift
-        install_extra_args=( "$@" )
-        install_script_dir=
-    fi
-    declare first rest
-    if [[ ! ${install_repo:-/} =~ / ]]; then
-        if [[ $install_repo =~ \.sh$ ]]; then
-            install_url ''
-            install_script_eval "$install_repo"
-            return
-        fi
-        first=download
-        rest=installers/$install_repo
-    elif [[ $install_repo =~ ^/*([^/].*[^/])/*$ ]]; then
-        first=${BASH_REMATCH[1]}
-        if [[ $first =~ ^([^/]+/[^/]+)/(.+)$ ]]; then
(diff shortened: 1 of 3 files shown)


bin/install.sh
```diff
@@ -248,41 +248,6 @@ install_os_release_vars() {
     fi
 }
 
-install_repo() {
```
robnagler, line 251: Since this is an interface change, I suggest you leave a routine like this:

install_repo() {
     install_error 'install_repo is deprecated, use install_repo_eval'
}

robnagler (changes requested):

bin/install.sh
rorour: 2d530e0aa135d083036203f79898575570e15441

robnagler (approved):

rorour:
@robnagler not sure if you saw the comment above, could you make the change to [radiasoft/container-python3](https://github.com/radiasoft/container-python3) or give me permission to push changes?

robnagler:
Sorry, missed that. Added you.

Issue #357, opened 2022-10-27
20221115 add ndiff to rpm-code/codes and add to beamsim-codes


robnagler:
It's small when compiled: https://www.math.utah.edu/~beebe/software/ndiff/

Needed for https://github.com/radiasoft/pykern/issues/311

gurhar1133:
@robnagler this is closed right?

PR #374, opened 2022-11-14
#357 Fix: ndiff rpm

gurhar1133:
ndiff added to beamsim-codes as well as an installer for rpm-code


e-carlin (approved):

e-carlin:
Should this be built for fedora 32?

PR #387, opened 2022-12-02
Fix #379: install h5ImageGenerator

### LINKED ISSUE #379: 20221130 h5ImageGenerator needed for ml
the h5imagegenerator python package is needed for [activait image preprocessing](https://github.com/radiasoft/sirepo/issues/4995) (memory efficient batch shuffling and scaling) 

@e-carlin  brought to my attention that adding to ml installer stuff won't be an option until we have moved to fedora 36. This could take weeks.

@moellep do we want to merge [#4995](https://github.com/radiasoft/sirepo/issues/4995) sooner than that? and if so do you know of any workarounds to merge/get ci checks passing @robnagler

e-carlin:
Use conditional so it works on f32 and f36

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -436,6 +436,14 @@ install_vars_export() {
     done
 }
 
+# TODO(e-carlin): remove when f36 is deployed (https://github.com/radiasoft/download/issues/379)
+install_version_fedora_is_36() {
+    if [[ $install_version_fedora = 36 ]]; then
+        return 0
+    fi
+    return 1
+}
+
 install_virt_vars() {
     export install_virt_docker=
     export install_virt_virtualbox=
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -436,9 +436,8 @@ install_vars_export() {
     done
 }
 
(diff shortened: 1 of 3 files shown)


gurhar1133 (approved):
Looks good. Is there some way I should test this?

bin/install.sh
```diff
@@ -436,6 +436,14 @@ install_vars_export() {
     done
 }
 
+# TODO(e-carlin): remove when f36 is deployed (https://github.com/radiasoft/download/issues/379)
+install_version_fedora_is_36() {
```
robnagler, line 440: When backporting, it's usually better to make the older version, e.g. 

```sh
install_version_fedora_lt_36() {
    if (( install_version_fedora < 36 )); then
        return 0
    fi
    return 1
}
```

You don't have to remember to take this code out when going to 38, for example.

robnagler (approved):

robnagler:
I'll let you merge.

bin/install.sh
e-carlin: a6bfa33

PR #390, opened 2022-12-05
Fix #381: add synergia back in for f32

e-carlin:
Also, wrapt had been previously moved into common.sh (during f36 migration). But, when f32 ml.sh was brought back it made its way back in. It can live in common and both versions of tensorflow (f32 and f36) depend on the same version of wrapt so no conditional is needed.


robnagler (commented):
lgtm

PR #392, opened 2022-12-07
fix #391: conditionally add synergia for f32

### LINKED ISSUE #391: Use conditional for synergia
In https://github.com/radiasoft/download/pull/390 I added synergia back into the list. But, it needs to be behind a conditional so it isn't included in f36.

e-carlin:


robnagler (commented):
lgtm

PR #393, opened 2022-12-16
fix pykern#329: ci defaults to only look in <pkg> and tests

rorour:
depends on https://github.com/radiasoft/pykern/pull/330


installers/python-ci/radiasoft-download.sh
```diff
@@ -51,7 +51,7 @@ python_ci_main() {
             if [[ -f test.sh ]]; then
                 bash test.sh
             else
-                pykern ci run
+                pykern ci run --github-ci-repo=$GITHUB_REPOSITORY
```
robnagler, line 54: This should just use the environment variable and not need an extra command line arg.

PR #399, opened 2023-01-12
Fix #398: Switch to python 3.9 in fedora 36

### LINKED ISSUE #398: Switch to python 3.9
In the fedora 36 / python 3.10 saga we've run into trouble running srw on python 3.10 (https://github.com/radiasoft/sirepo/issues/5276). It works on 3.9 so were going to see if we can get all codes working on 3.9

e-carlin:
We tried upgrading to python 3.10 in the fedora 36 migration. But we ran into issues with it and srw
(https://github.com/radiasoft/sirepo/issues/5276).  Python 3.9 works and will be supported for some time so switching to it.


e-carlin:
Merging so I can get a test vm going but please review.

PR #403, opened 2023-01-16
fix #402: declare on _rsbeams_codes causes unbound variable

### LINKED ISSUE #402: fedora 36: no rs* packages installed in rscode-rsbeams

e-carlin:
This is some bit of bash that I don't understand, was unable to reproduce in a local script, and can't find an answer for online. @robnagler when _rsbeam_codes is defined with declare then the array is empty. Likewise calling `${#_rsbeam_codes[@]}` results in unbound variable. Removing the `declare` fixed the problem. Do you have any idea why that might be?

### DIFF
--- a/installers/rpm-code/codes/rsbeams.sh
+++ b/installers/rpm-code/codes/rsbeams.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-declare -a _rsbeam_codes=(
+_rsbeam_codes=(
     rsbeams
     rsflash
     rslaser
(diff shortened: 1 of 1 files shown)


robnagler:
>  Removing the `declare` fixed the problem. 

```sh
declare x
echo ${x-unset}
declare x=
echo ${x+set}
```

produces:

```txt
unset
set
```

The language in the various manuals is loose. `declare` is for typing. It can also be used to `set` a value. Without the `=` it just identifies a type associated with the name, but the variable is not set.

robnagler (approved):

installers/rpm-code/codes/rsbeams.sh
```diff
@@ -12,7 +12,7 @@ declare -a _rsbeam_codes=(
 rsbeams_main() {
     rsbeams_init_vars
     codes_dependencies common ml
-    local r
+    declare r
     for r in "${_rsbeam_codes[@]}"; do
```
robnagler, line 16: If you want to protect this, you have to: `${_rsbeam_codes[@]+"${_rsbeam_codes[@]}"}`

PR #409, opened 2023-01-31
Fix #408 PYKERN_PKCLI_TEST_MAX_FAILURES=1

### LINKED ISSUE #408: 2022215 config test.max_failures = 1 in github action
If there's a failure, bail the action. There's almost no value in seeing other failures.

robnagler:


rorour (commented):
Looks good to me. 
It seems like tests have been randomly failing more often recently (last month or so). Especially nginx_uwsgi_test.py. Do you know what's causing that?

robnagler:
No, Evan, Paul, and I were just discussing that. With getting rid of Flask, this particular test won't be an issue. I suspect it is either nginx or uwsgi (probably latter) that is doing something in the GitHub Action environment that isn't "correct".

PR #437, opened 2023-03-20
fix #435: ci builds with latest pykern and sirepo

### LINKED ISSUE #435: ci builds with latest pykern and sirepo
We want ci builds with latest pykern and sirepo even if not in synch with alpha for repos besides sirepo where we us ci

rorour:


installers/python-ci/radiasoft-download.sh
```diff
@@ -8,6 +8,7 @@ python_ci_main() {
     fi
     declare i
     declare p=0
+    declare s=1
```
robnagler, line 11: Default should be `s=0`.  Only case that needs `s=1` is the default (`*)`).

With two packages, use an array, e.g.

```sh
declare -a p=()
```

Then add to it below:

```sh
p+=( pykern sirepo )
```

installers/python-ci/radiasoft-download.sh
```diff
@@ -43,9 +45,13 @@ python_ci_main() {
             set -x
             cd '$d'
             export GITHUB_TOKEN='${GITHUB_TOKEN:-}'
-            if (( $p )); then
+            if (( $p )) || (( $s )); then
```
robnagler, line 48: `$p` and `$s` do not need to be coupled. They are implicitly coupled because pykern is installed before sirepo.

Using the array above, this becomes simpler:

```sh
for x in ${p[*]}; do
    pip uninstall -y \$x >& /dev/null || true
    pip install git+https://github.com/radiasoft/\$x.git
done
```

There are a couple of assumptions that are safe, since they are localized:
- `$p` contains words without spaces (Python packages cannot have spaces)
- `$p` is properly ordered (pykern comes before sirepo)

robnagler (changes requested):

installers/python-ci/radiasoft-download.sh
rorour: d86f234e9014fadfaad2084d990b05c3a43068a4

rorour:
@robnagler ping

installers/python-ci/radiasoft-download.sh
```diff
@@ -43,10 +44,11 @@ python_ci_main() {
             set -x
             cd '$d'
             export GITHUB_TOKEN='${GITHUB_TOKEN:-}'
-            if (( $p )); then
-                pip uninstall -y pykern >& /dev/null || true
-                pip install git+https://github.com/radiasoft/pykern.git
-            fi
+            for x in "${p[*]}"
```
robnagler, line 47: Two syntax issues:

`for x in ${p[*]}; do`

The quotes are unnecessary, and I believe will cause issues. This assumes $p values do not contain spaces or specials (which is true), but it should be noted.

installers/python-ci/radiasoft-download.sh
```diff
@@ -43,10 +44,11 @@ python_ci_main() {
...
+            do
+                pip uninstall -y \$x >& /dev/null || true
+                pip install git+https://github.com/radiasoft/\$x.git
```
robnagler, line 50: I think we will want the this:

`pip install git+https://'${GITHUB_TOKEN:+$GITHUB_TOKEN@}'/radiasoft/\$x.git`

The GITHUB_TOKEN avoids rate limiting issues we have run into in Actions.

installers/python-ci/radiasoft-download.sh
```diff
@@ -43,10 +44,11 @@ python_ci_main() {
             set -x
```
robnagler, line 44: This is on line 42 (`su - vagrant`). The EOF2 should be in single quotes because `\$x` will get interpolated twice, and we only want it to be once.

robnagler (changes requested):

installers/python-ci/radiasoft-download.sh
rorour: 085af5b7bc7474e434c3f7ad7cc88ba32ed8cbb8

installers/python-ci/radiasoft-download.sh
rorour: 085af5b7bc7474e434c3f7ad7cc88ba32ed8cbb8

installers/python-ci/radiasoft-download.sh
rorour: 085af5b7bc7474e434c3f7ad7cc88ba32ed8cbb8

installers/python-ci/radiasoft-download.sh
```diff
@@ -38,15 +39,16 @@ python_ci_main() {
...
             set -eou pipefail
-            set -x
             cd '$d'
             export GITHUB_TOKEN='${GITHUB_TOKEN:-}'
-            if (( $p )); then
-                pip uninstall -y pykern >& /dev/null || true
-                pip install git+https://github.com/radiasoft/pykern.git
-            fi
+            # POSIT: no spaces or specials in $p values
+            for x in ${p[*]}
```
robnagler, line 47: RS bash style is `for x in y; do` not `for x in y
do`.

installers/python-ci/radiasoft-download.sh
```diff
@@ -38,15 +39,16 @@ python_ci_main() {
...
+            do
+                pip uninstall -y \$x >& /dev/null || true
+                pip install git+https://'${GITHUB_TOKEN:+$GITHUB_TOKEN@}'/radiasoft/\$x.git
```
robnagler, line 50: Missing github.com. I apologize for giving the incorrect URL in the comment.

robnagler (changes requested):

installers/python-ci/radiasoft-download.sh
rorour: 3d91f17e1d051fa7340c1c5aafd7819b1e38da6d

robnagler (approved):

PR #444, opened 2023-03-31
Fix #350: Fix nfs mounts for fedora 36.

### LINKED ISSUE #350: 20221215 fedora36/monterey vagrant-dev nfs mount is failing
Workaround: `vagrant_dev_no_mounts=1 radia_run vagrant-dev fedora`
```sh
==> default: Mounting NFS shared folders...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!
mount -o vers=3,nolock,fsc,actimeo=2 10.10.10.1:/System/Volumes/Data/Users/<user>/v2 /vagrant
Stdout from the command:
Stderr from the command:
mount.nfs: an incorrect mount option was specified
```

I tried disabling selinux and mounting simply:

```sh
mount -t nfs 10.10.10.1:/System/Volumes/Data/Users/<user>/v2 /vagrant
mount.nfs: Broken pipe
```

e-carlin:
The root cause of the problems is that vagrant doesn't support NetworkManager which is used in Fedora 36
https://github.com/hashicorp/vagrant/issues/12762. So, the private_network we define in the Vagrantfile gets setup on the host but not in the guest. We fix this by using vagrant_dev_provision_eth1 (set the interface ourselves). But, this provisioner runs after the sync_folder(s) are created. So, the synced_folders fail when using nfs because networking isn't setup right.

There is no way to run a provisioner before a synced_folder, https://github.com/hashicorp/vagrant/issues/936. So, we must set `disabled: true`, bring the vm up, set `disabled: false`, and reload the vm.

In addition, git introduced checking of owners for directories https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9. In the NFS case the UID/GID from the host likely doesn't match the UID/GID of the user in the vm. This causes git to complain. So, we need to set safe.directory to '*'. This configuration is set inside of the vm and we only have one user inside of vms so it is safe.

### DIFF
--- a/installers/home/radiasoft-download.sh
+++ b/installers/home/radiasoft-download.sh
@@ -4,5 +4,6 @@
 #
 home_main() {
     install_url biviosoftware/home-env
-    install_script_eval install.sh
+    bivio_home_env_ignore_git_dir_ownership=${vagrant_dev_ignore_git_dir_ownership:+1}  \
+        install_script_eval install.sh
 }
--- a/installers/home/radiasoft-download.sh
+++ b/installers/home/radiasoft-download.sh
@@ -4,6 +4,5 @@
 #
 home_main() {
     install_url biviosoftware/home-env
-    bivio_home_env_ignore_git_dir_ownership=${vagrant_dev_ignore_git_dir_ownership:+1}  \
-        install_script_eval install.sh
+    install_script_eval install.sh
 }
--- a/installers/vagrant-dev/radiasoft-download.sh
+++ b/installers/vagrant-dev/radiasoft-download.sh
@@ -59,6 +59,12 @@ vagrant_dev_first_up() {
     fi
 }
 
+vagrant_dev_ignore_git_dir_ownership() {
+    declare os="$1"
+    if [[ ! ${vagrant_dev_no_nfs_src:+1} && $os =~ fedora && ! $(install_version_fedora_lt_36) ]]; then
+        echo 1
+    fi
+}
 
 vagrant_dev_ip() {
     declare host=$1
@@ -181,7 +187,9 @@ expects: fedora|centos[/<version>], <ip address>, update, v[1-9].radia.run"
     fi
     vagrant ssh <<EOF
 $(install_vars_export)
-curl $(install_depot_server)/index.sh | bash -s redhat-dev "${vagrant_dev_no_nfs_src:+1}"
+curl $(install_depot_server)/index.sh | \
+  vagrant_dev_ignore_git_dir_ownership=$(vagrant_dev_ignore_git_dir_ownership $os) \
+  bash -s redhat-dev
 EOF
     vagrant_dev_post_install
 }
--- a/installers/vagrant-dev/radiasoft-download.sh
+++ b/installers/vagrant-dev/radiasoft-download.sh
@@ -49,7 +49,7 @@ vagrant_dev_first_up() {
     declare host="$2"
     declare ip="$3"
     if [[ ! ${vagrant_dev_no_vbguest:+1} ||  ! ${vagrant_dev_no_mounts:+1} && ${vagrant_dev_provision_eth1:+1} ]]; then
-        vagrant_dev_vagrantfile "$os" "$host" "$ip" '1'
+        vagrant_dev_vagrantfile "$os" "$host" "$ip" 1
         vagrant up
         # TODO(e-carlin):  uncomment
 #         vagrant ssh <<'EOF'
@@ -188,7 +188,7 @@ expects: fedora|centos[/<version>], <ip address>, update, v[1-9].radia.run"
     vagrant ssh <<EOF
 $(install_vars_export)
 curl $(install_depot_server)/index.sh | \
-  vagrant_dev_ignore_git_dir_ownership=$(vagrant_dev_ignore_git_dir_ownership $os) \
+  bivio_home_env_ignore_git_dir_ownership=$(vagrant_dev_ignore_git_dir_ownership $os) \
   bash -s redhat-dev
 EOF
     vagrant_dev_post_install
@@ -200,9 +200,9 @@ vagrant_dev_mounts() {
         echo 'config.vm.synced_folder ".", "/vagrant", disabled: true'
         return
     fi
-    declare d='false'
+    declare d=false
     if [[ $first ]]; then
-        d='true'
+        d=true
     fi
     # Have to use proto=tcp otherwise mount defaults to udp which doesn't work in f36
     declare f=' type: "nfs", mount_options: ["nolock", "fsc", "actimeo=2", "proto=tcp"], nfs_udp: false, disabled: '"$d"
--- a/installers/vagrant-dev/radiasoft-download.sh
+++ b/installers/vagrant-dev/radiasoft-download.sh
@@ -44,6 +44,14 @@ vagrant_dev_box_add() {
     fi
 }
 
+vagrant_dev_enable_mounts() {
+    if [[ ${vagrant_dev_no_mounts:+1} ]]; then
+        return
+    fi
+    perl -pi -e 's/(synced_folder.*nfs.*disabled: )true(.*)/$1false$2/' Vagrantfile
+    vagrant reload
+}
+
 vagrant_dev_ip() {
     declare host=$1
     declare i=$(dig +short "$host" 2>/dev/null || true)
@@ -129,9 +137,10 @@ expects: fedora|centos[/<version>], <ip address>, update, v[1-9].radia.run"
         vagrant_dev_no_nfs_src=1
     fi
 #TODO(robnagler) handle fedora/<version> syntax
-    if [[ ! ${vagrant_dev_no_mounts+1} && $os =~ fedora ]] && (( $install_version_fedora >= 36 )); then
-        vagrant_dev_no_mounts=1
+    if [[ ! ${vagrant_dev_provision_eth1+1} && $os =~ fedora && ! $(install_version_fedora_lt_36) ]]; then
         vagrant_dev_no_vbguest=${vagrant_dev_no_vbguest-}
+        # Vagrant doesn't handle NetworkManager correctly so must handle ourselves
+        # https://github.com/hashicorp/vagrant/issues/12762
         vagrant_dev_provision_eth1=1
     fi
     if [[ $_vagrant_dev_host_os == ubuntu ]]; then
@@ -156,6 +165,7 @@ EOF
     fi
     vagrant_dev_vagrantfile "$os" "$host" "$ip" ''
     vagrant up
+    vagrant_dev_enable_mounts
     if [[ ${vagrant_dev_no_dev_env:+1} ]]; then
         return
     fi
@@ -171,7 +181,7 @@ EOF
     fi
     vagrant ssh <<EOF
 $(install_vars_export)
-curl $(install_depot_server)/index.sh | bash -s redhat-dev
+curl $(install_depot_server)/index.sh | bash -s redhat-dev "${vagrant_dev_no_nfs_src:+1}"
 EOF
     vagrant_dev_post_install
 }
@@ -181,8 +191,10 @@ vagrant_dev_mounts() {
         echo 'config.vm.synced_folder ".", "/vagrant", disabled: true'
         return
     fi
-    # Have to use vers=3 b/c vagrant will insert it (incorrectly) otherwise. Not sure why.
-    declare f=' type: "nfs", mount_options: ["nolock", "fsc", "actimeo=2"], nfs_version: 3, nfs_udp: false'
+    # Have to use proto=tcp otherwise mount defaults to udp which doesn't work in f36
+    # disabled: true because vagrant_dev_provision_eth1 needs to run before nfs mount will work.
+    # vagrant_dev_enable_mounts sets disabled: false
+    declare f=' type: "nfs", mount_options: ["nolock", "fsc", "actimeo=2", "proto=tcp"], nfs_udp: false, disabled: true'
     declare res=( 'config.vm.synced_folder ".", "/vagrant",'"$f" )
     if [[ ! ${vagrant_dev_no_nfs_src:+1} ]]; then
         mkdir -p "$HOME/src"
--- a/installers/vagrant-dev/radiasoft-download.sh
+++ b/installers/vagrant-dev/radiasoft-download.sh
@@ -44,13 +44,21 @@ vagrant_dev_box_add() {
     fi
 }
 
-vagrant_dev_enable_mounts() {
-    if [[ ${vagrant_dev_no_mounts:+1} ]]; then
-        return
+vagrant_dev_first_up() {
+    declare os="$1"
+    declare host="$2"
+    declare ip="$3"
+    if [[ ! ${vagrant_dev_no_vbguest:+1} ||  ! ${vagrant_dev_no_mounts:+1} && ${vagrant_dev_provision_eth1:+1} ]]; then
+        vagrant_dev_vagrantfile "$os" "$host" "$ip" '1'
+        vagrant up
+        # TODO(e-carlin):  uncomment
+#         vagrant ssh <<'EOF'
+# sudo yum install -q -y kernel kernel-devel kernel-headers kernel-tools perl
+# EOF
+        vagrant halt
     fi
-    perl -pi -e 's/(synced_folder.*nfs.*disabled: )true(.*)/$1false$2/' Vagrantfile
-    vagrant reload
 }
+
 
 vagrant_dev_ip() {
     declare host=$1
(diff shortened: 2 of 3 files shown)


installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -44,6 +44,14 @@ vagrant_dev_box_add() {
     fi
 }
 
+vagrant_dev_enable_mounts() {
+    if [[ ${vagrant_dev_no_mounts:+1} ]]; then
+        return
+    fi
+    perl -pi -e 's/(synced_folder.*nfs.*disabled: )true(.*)/$1false$2/' Vagrantfile
```
robnagler, line 51: vagrant_dev_vagrantfile is parameterized. If $first is set, disable mounts. This currently only done for vbguest, but now it needs to be done for this case, too. Don't edit a file that's being generated by the same script, iow.

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -171,7 +181,7 @@ EOF
     fi
     vagrant ssh <<EOF
 $(install_vars_export)
-curl $(install_depot_server)/index.sh | bash -s redhat-dev
+curl $(install_depot_server)/index.sh | bash -s redhat-dev "${vagrant_dev_no_nfs_src:+1}"
```
robnagler, line 184: When crossing script boundaries use variables, not anonymous args (home-env/install.sh was a mistake). `redhat_dev_git_nfs` would be the variable, but I don't think this should run here. This is a vagrant-dev issue. It will not happen in docker which also uses redhat-dev. 

This doesn't allow a cli override. Do it like the other args, and maybe inside vagrant_dev_main.

I would define it is `vagrant_dev_git_nfs` (affirmative) so it would default to true unless set in the environment. This might need a version check on the git as this wasn't always there so that it works on centos 7.

robnagler (changes requested):

installers/vagrant-dev/radiasoft-download.sh
e-carlin: Can you explain the code a bit more? I understand that I can call `vagrant_dev_vagrantfile` again instead of editing it as Im doing here. I'll make that change. What I don't understand is what `$first` has to do with the mounts? From my reading of the code `$first` is gated by `vagrant_dev_no_vbguest` which I don't believe has any relationship with nfs mounts. What am I not seeing?

installers/vagrant-dev/radiasoft-download.sh
e-carlin: > When crossing script boundaries use variables, not anonymous args

Will do.

>  but I don't think this should run here. This is a vagrant-dev issue

The problem is it needs to run after we have git installed (comes from redhat-base) but before we do any git operations in ~/src (ex `install_repo_eval home`). So, it can't be run in vagrant-dev w/o installing git in vagrant-dev and running before redhat-dev. Or breaking up redhat-dev and calling the pieces individually in vagrant-dev. What do you think would be best?

> It will not happen in docker which also uses redhat-dev.

Can you show me where that is? I searched and couldn't find the use https://github.com/search?type=code&auto_enroll=true&q=org%3Aradiasoft+redhat-dev

installers/vagrant-dev/radiasoft-download.sh
robnagler: $first means "first time through do something different". This is necessary for vbguest and now it is important for this case. The logic might look something like:

```sh
    if [[ ! ${vagrant_dev_no_vbguest:+1} ||  ! ${vagrant_dev_no_mounts:+1} && $has_network_manager ]]; then
        vagrant_dev_vagrantfile "$os" "$host" "$ip" '1'
        vagrant up
        if [[ ${vagrant_dev_no_vbguest:+1} ]]; then
            vagrant ssh <<'EOF'
sudo yum install -q -y kernel kernel-devel kernel-headers kernel-tools perl
EOF
        fi
        vagrant halt
    fi
```

It might be argued that this code and the subsequent few lines might be moved to `vagrant_dev_up()`, a new function.

There's an important point here on the UI. If vbguest and mounts are supplied (the default mode we would actually like to have), then there would be three boots instead of one with the existing code, and the above code does the same thing with two boots.

installers/vagrant-dev/radiasoft-download.sh
robnagler: > The problem is it needs to run after we have git installed 

This is subtle in a number of ways.

There is a conflict with install_repo_eval, which installs `~/.git_config` only if it doesn't already exist. However, it will exist as soon as `git config --global --add safe.directory '*'` is called.

I think home-env actually is ok, because all it does is clone, which doesn't need safe.directory.  This works:

```sh
sudo chmod 775 .
sudo chown root .
git clone https://github.com/biviosoftware/home-env
```

This does not work:

```sh
cd home-env
sudo chmod 775 .
sudo chown root .
git pull
fatal: detected dubious ownership in repository at '/home/vagrant/src/biviosoftware/home-env'
```

So, I think we are ok to leave it in vagrant-dev. Once vagrant-dev happens, it will be configured.

For idempotency, you could call it twice inside vagrant-dev: before redhat-dev and after. The function should check for the existence of `~/.gitconfig` and do nothing if it is not there.

> # POSIT: redhat-base installs git
> redhat_dev_fix_git_nfs_permissions ${1:+1}

This does not need to be posited. POSIT is only for cases where the assumption is not asserted by the code. `git` is called in redhat_dev_fix_git_nfs_permissions so it will fail immediately.

Bash functions should declare their arguments so the use of $1 should occur in the first line of this function.

You should document why this is necessary and how it couples with vagrant-dev.

Does the fix need to happen for the root user's gitconfig? I don't think that's right, because it really is a problem if root is in an arbitrary directory so I would say no, and I would document it as such.

While it is problematic for vagrant as well, it should be clearly documented that we consider it safe since this is a development VM. If the VM is compromised, it doesn't need to be done via git for a VM that has no sudo password for the user vagrant...

e-carlin: > There is a conflict with install_repo_eval, which installs ~/.git_config

Yes, I meant to leave a comment about this. That is why I do the add after [the copy from the host](https://github.com/radiasoft/download/blob/a2e71aaa4e4f5238b448d6e85d1cebff269975c6/installers/vagrant-dev/radiasoft-download.sh#L175). But, I couldn't find an easy way to resolve with home-env...

> I think home-env actually is ok, because all it does is clone, which doesn't need safe.directory

I have ~/src/biviosoftware/home-env on my host machine. It get's mounted into the vm. So install.sh [does a pull](https://github.com/biviosoftware/home-env/blob/06c2f9b7bd1a6e54823df616406e0f7bfbcee0b9/install.sh#L45) which runs into the issue.

> For idempotency, you could call it twice inside vagrant-dev: before redhat-dev and after.

There is a lot going on here so I could be lost. But, git doesn't exist before redhat-dev so I don't think we can call it?

> Does the fix need to happen for the root user's gitconfig?

We don't mount anything under /root so we won't run into any problems. I'll make a note.

> it doesn't need to be done via git for a VM that has no sudo password for the user vagrant

lol, agreed.

robnagler: We had a conversation about this. I think we have a plan now.

installers/vagrant-dev/radiasoft-download.sh
e-carlin: Thanks. My myopia couldn't see past that I could still check vagrant_dev_* variables along with $first.

6c32e5c

installers/vagrant-dev/radiasoft-download.sh
e-carlin: a5ce25c

e-carlin:
Ready for another review.

installers/home/radiasoft-download.sh
```diff
@@ -4,5 +4,6 @@
 #
 home_main() {
     install_url biviosoftware/home-env
-    install_script_eval install.sh
+    bivio_home_env_ignore_git_dir_ownership=${vagrant_dev_ignore_git_dir_ownership:+1}  \
```
robnagler, line 7: That's not usually how we cascade, but I guess it works. Usually it's a pure value copy:

```sh
bivio_home_env_ignore_git_dir_ownership=${vagrant_dev_ignore_git_dir_ownership:-}
```

I wonder if the cascade is necessary (see below).

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -171,18 +187,25 @@ EOF
     fi
     vagrant ssh <<EOF
 $(install_vars_export)
-curl $(install_depot_server)/index.sh | bash -s redhat-dev
+curl $(install_depot_server)/index.sh | \
+  vagrant_dev_ignore_git_dir_ownership=$(vagrant_dev_ignore_git_dir_ownership $os) \
```
robnagler, line 191: Typically we would name the variable at the place where it is used, e.g. you would set `bivio_home_env_ignore_git_dir_ownership`. This presumes that redhat-dev cascades the value, which it might need to(?).

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -171,18 +187,25 @@ EOF
...
+  bash -s redhat-dev
 EOF
     vagrant_dev_post_install
 }
 
 vagrant_dev_mounts() {
+    declare first="${1:+1}"
```
robnagler, line 198: I don't think the quotes are necessary. And I would think we'd want to abort if the parameter wasn't passed. This is an internal function.

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -44,6 +44,28 @@ vagrant_dev_box_add() {
     fi
 }
 
+vagrant_dev_first_up() {
+    declare os="$1"
```
robnagler, line 48: I don't think quotes are necessary

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -44,6 +44,28 @@ vagrant_dev_box_add() {
...
+    declare host="$2"
+    declare ip="$3"
+    if [[ ! ${vagrant_dev_no_vbguest:+1} ||  ! ${vagrant_dev_no_mounts:+1} && ${vagrant_dev_provision_eth1:+1} ]]; then
+        vagrant_dev_vagrantfile "$os" "$host" "$ip" '1'
```
robnagler, line 52: quotes around the 1 are not necessary.

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -171,18 +187,25 @@ EOF
...
 
 vagrant_dev_mounts() {
+    declare first="${1:+1}"
     if [[ ${vagrant_dev_no_mounts:+1} ]]; then
         echo 'config.vm.synced_folder ".", "/vagrant", disabled: true'
         return
     fi
-    # Have to use vers=3 b/c vagrant will insert it (incorrectly) otherwise. Not sure why.
-    declare f=' type: "nfs", mount_options: ["nolock", "fsc", "actimeo=2"], nfs_version: 3, nfs_udp: false'
+    declare d='false'
```
robnagler, line 203: no quotes

robnagler (changes requested):

installers/vagrant-dev/radiasoft-download.sh
e-carlin: Can you elaborate on why no to use quotes? I get that the value in $1 has no spaces in it so quotes aren't strictly necessary. But, in my mind that creates more tightly coupled code. We are coupling the fact that the value has no quotes to how we are using it here and that coupling isn't necessary. 

Somewhere along the way we need to set a default for $first. Otherwise bash will complain about an unbound variable.

```bash
a() {
    b '1'
    b ''
}

b() {
    declare first=$1
    x=$(c $first)
    echo "$x"
}

c() {
    declare first=$1
    echo "first=$first"
}

a
```
```
$ bash x.sh
first=1
x.sh: line 15: $1: unbound variable
```
 
So should we set the default when calling? Is there a reason to prefer handling default in the caller vs the callee?

installers/home/radiasoft-download.sh
e-carlin: 9b21050

installers/vagrant-dev/radiasoft-download.sh
e-carlin: 9b21050

installers/vagrant-dev/radiasoft-download.sh
e-carlin: 9b21050

installers/vagrant-dev/radiasoft-download.sh
e-carlin: 9b21050

installers/vagrant-dev/radiasoft-download.sh
e-carlin: Same question as https://github.com/radiasoft/download/pull/444#discussion_r1157801266

installers/vagrant-dev/radiasoft-download.sh
robnagler: The error that causes the unbound variable is on line 8, it should be:

```sh
    x=$(c "$first")
```

Then, the script runs correctly.  You need to quote in command expansions (`c $first`) and array values, e.g. 

```sh
$ x=( $(seq 3) ) y=( "$(seq 3)" )
$ (IFS=,; echo "${x[*]}" "${y[*]}")
1,2,3 1
2
3
```

This is the only expansion that happens in assignments:

> Each variable assignment is checked for unquoted tilde-prefixes immediately following a ":" or "=". In these cases, tilde expansion is also performed. Consequently, one may use file names with tildes in assignments to PATH, MAILPATH, and CDPATH, and the shell assigns the expanded value.

I asked ChatGPT, and it seems it didn't know this. It does

installers/vagrant-dev/radiasoft-download.sh
robnagler: I would say one thing: we need better tests to give confidence that things are working correctly. In Bash, it's pretty easy to mock things, e.g. add a command in the path that gets used instead of `vagrant`, for example.

In an ideal world, we'd program everything in a type safe language, and it would be as easy to develop as it is with bash, but that world doesn't exist. The nature of shells is that they are untyped. 

That's why I think testing is the only way to convince ourselves that the program is running correctly.

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -355,7 +377,7 @@ vagrant_dev_vagrantfile() {
     # vagrant_dev_box_add returns in box
     declare box
     vagrant_dev_box_add "$os"
-    declare mounts="$(vagrant_dev_mounts)"
+    declare mounts="$(vagrant_dev_mounts $first)"
```
robnagler, line 380: The correct syntax is:

```sh
    declare mounts=$(vagrant_dev_mounts "$first")
```

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -171,18 +186,25 @@ EOF
     fi
     vagrant ssh <<EOF
 $(install_vars_export)
-curl $(install_depot_server)/index.sh | bash -s redhat-dev
+curl $(install_depot_server)/index.sh | \
```
robnagler, line 189: Quote `"$os"`. 

Ideally, we should encapsulate the curl, index.sh in a function that is something like:

```sh
export bivio_home_env_ignore_git_dir_ownership=$(vagrant_dev_ignore_git_dir_ownership $os) 
$(install_repo_in_subshell redhat-dev)
```

This would implicitly set install_vars_export, and do the curl appropriately, possibly also setting `-euo`. For another day, but noting that there's a lot of code here, unnecessarily.

I like the export on its own line since it makes the code less messy.

installers/vagrant-dev/radiasoft-download.sh
e-carlin: Thanks.

If I wanted to quote everything is this right? `x="$(c "$first")"`?

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -171,18 +186,25 @@ EOF
...
 
 vagrant_dev_mounts() {
+    declare first="${1:+1}"
     if [[ ${vagrant_dev_no_mounts:+1} ]]; then
         echo 'config.vm.synced_folder ".", "/vagrant", disabled: true'
         return
     fi
-    # Have to use vers=3 b/c vagrant will insert it (incorrectly) otherwise. Not sure why.
-    declare f=' type: "nfs", mount_options: ["nolock", "fsc", "actimeo=2"], nfs_version: 3, nfs_udp: false'
+    declare d=false
```
robnagler, line 202: I particularly dislike this construct, but there's nothing really better. People try to write [some pretty crazy stuff](https://stackoverflow.com/q/3953645), but in the end, this is probably the most straight forward. However, I just thought of this one, which I haven't seen used elsewhere:

```sh
declare d=${first:+true}
: ${d:=false}
```

robnagler (changes requested):

installers/vagrant-dev/radiasoft-download.sh
robnagler: Yes, that's the right way to do it, and typically used elsewhere, e.g.

```sh
echo "$(c "$first")"
```

The quotes inside of subshell expansions is why subshell expansions were introduced. This doesn't work in backticks.

e-carlin: Thanks for walking me through it b2d113e

installers/vagrant-dev/radiasoft-download.sh
e-carlin: b2d113e

installers/vagrant-dev/radiasoft-download.sh
e-carlin: Agreed. Just the quote for now b2d113e

installers/vagrant-dev/radiasoft-download.sh
e-carlin: I don't have a strong opinion either way

robnagler (approved):

PR #446, opened 2023-03-31
Fix #445 protect $p expansion from empty array

### LINKED ISSUE #445: python-ci line 35: p: unbound variable

robnagler:
I'm going to push this because pykern can't build. I would still like a review.

### DIFF
--- a/installers/python-ci/radiasoft-download.sh
+++ b/installers/python-ci/radiasoft-download.sh
@@ -43,8 +43,9 @@ python_ci_main() {
             set -eou pipefail
             cd '$d'
             export GITHUB_TOKEN='${GITHUB_TOKEN:-}'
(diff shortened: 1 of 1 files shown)


rorour:
Was it causing an error before? I don't see any problems when I test the before code with a dummy script. And why declare x?

robnagler:
> Was it causing an error before? 

#445 doesn't describe it well. Pykern was failing, because `declare -a p=()` means `$p` is undefined (even though it exists in some way). It's an odd case, which isn't true for non-array variables. 

> I don't see any problems when I test the before code with a dummy script. And why declare x?

We declare all variables before referring to them. This isn't required by Bash, but it's nice to be formal in scripts. Declaring demonstrations intention.

PR #456, opened 2023-04-16
Fix #454: Support updating python version

### LINKED ISSUE #454: nersc-sirepo-update should check python version == $install_version_python
If they aren't equal it should update the python version. This will allow us to release f36 with py3.9.

Our script for running releases should be adapted to set install_version_python in the nersc-sirepo-update command printed out. This will allow us to cascade the python version from the build server to NERSC.

e-carlin:


installers/nersc-sirepo-update/radiasoft-download.sh
```diff
@@ -44,3 +42,19 @@ nersc_sirepo_update_main() {
...
 }
+
+nersc_sirepo_update_pyenv() {
+    virtualenv_name=$1
+    if ! pyenv versions --bare | grep -q "^$install_version_python$"; then
+        nersc_pyenv_no_global=1 install_repo_eval nersc-pyenv
+    fi
+    if [[ -e ~/.pyenv/versions/$virtualenv_name ]]; then
+        install_not_strict_cmd pyenv shell "$virtualenv_name"
+        if [[ $(python --version | cut -d' ' -f2) != $install_version_python ]]; then
```
robnagler, line 53: We should really add `nth` or equivalent to install, e.g. `$(install_nth 2 "$(python --version)")`, which abstracts the above, and doesn't use pipes.

robnagler (approved):

PR #472, opened 2023-05-01
Fix #467: fix mantid on f36

### LINKED ISSUE #467: f36 import mantid fails
```
CMake (main)$ rsmantid -c 'from mantid import simpleapi'
Traceback (most recent call last):
  File "/home/vagrant/.local/lib/mantid/lib/mantid/utils/_utils.py", line 37, in import_mantid_cext
    lib = import_module(modulename, package)
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 666, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 565, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 1173, in create_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
ImportError: /home/vagrant/.local/lib/mantid/lib/mantid/api/../../../lib/libMantidGeometry.so: undefined symbol: _ZN24BRepBuilderAPI_MakeShape5BuildEv

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/vagrant/.local/lib/mantid/lib/mantid/utils/_utils.py", line 41, in import_mantid_cext
    lib = import_module(modulename.lstrip('.'))
  File "/home/vagrant/.pyenv/versions/3.9.15/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 984, in _find_and_load_unlocked
ModuleNotFoundError: No module named '_api'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/vagrant/.local/lib/mantid/lib/mantid/__init__.py", line 109, in <module>
    from mantid import api as _api  # noqa: F401, E402
  File "/home/vagrant/.local/lib/mantid/lib/mantid/api/__init__.py", line 20, in <module>
    import_mantid_cext('._api', 'mantid.api', globals())
  File "/home/vagrant/.local/lib/mantid/lib/mantid/utils/_utils.py", line 46, in import_mantid_cext
    raise ImportError(msg)
ImportError: First import of "._api" failed with "/home/vagrant/.local/lib/mantid/lib/mantid/api/../../../lib/libMantidGeometry.so: undefined symbol: _ZN24BRepBuilderAPI_MakeShape5BuildEv". Second import of "_api" failed with "No module named '_api'"
```

I believe that error [comes from opencascade](https://forum.freecad.org/viewtopic.php?style=5&t=68128). Opencascade was added when we upgraded to f36 (and upgraded mantid versions). It has also caused problems by increasing the size of the beamsim image by 3gb. It is unclear to me, at the moment, what exactly opencascade is used for. But, my best guess is 3d visualization. Since we don't do any visualization with mantid (we just use the python api) I think it might be safe to remove it. There is a cmake option [to disable it](https://github.com/mantidproject/mantid/blob/v6.4.0/buildconfig/CMake/CommonSetup.cmake#L59) which I'll try.

e-carlin:
Remove opencascad-devel which brought in 3GB of dependencies. It caused the error on import, was large, and was unused because we don't do any rendering with Mantid. In addition, disable other unused parts of Mantid (ex docs).

Add ${module}_test which we can use for adding tests to modules which will hopefully prevent regressions like this from happening in the future.

### DIFF
--- a/installers/rpm-code/codes.sh
+++ b/installers/rpm-code/codes.sh
@@ -5,7 +5,7 @@ codes_assert_easy_install() {
         # local environment may have easy-install.pth
         return
     fi
-    local easy=$(find "${codes_dir[pyenv_prefix]}"/lib -name easy-install.pth)
+    declare easy=$(find "${codes_dir[pyenv_prefix]}"/lib -name easy-install.pth)
     if [[ $easy ]]; then
         install_err "$easy: packages used python setup.py install instead of pip:
 $(cat "$easy")"
@@ -15,7 +15,7 @@ $(cat "$easy")"
 codes_cmake() {
     mkdir build
     cd build
-    local t=Release
+    declare t=Release
     if [[ ${CODES_DEBUG_FLAG:-} ]]; then
         t=Debug
     fi
@@ -23,7 +23,7 @@ codes_cmake() {
 }
 
 codes_cmake_build() {
-    local cmd=( cmake --build . -j$(codes_num_cores) )
+    declare cmd=( cmake --build . -j$(codes_num_cores) )
     if [[ ${CODES_DEBUG_FLAG:-} ]]; then
         cmd+=( --verbose )
     fi
@@ -48,8 +48,8 @@ codes_dependencies() {
 }
 
 codes_dir_setup() {
-    local todo=()
-    local d n p
+    declare todo=()
+    declare d n p
     for n in prefix \
         etc \
         bashrc_d \
@@ -77,6 +77,7 @@ codes_dir_setup() {
         install_msg 'creating directories'
         mkdir -p "${todo[@]}"
     else
+
         # common doesn't use pyenv_prefix, it creates it
         codes_dir[pyenv_prefix]=$(realpath "$(pyenv prefix)")
     fi
@@ -84,10 +85,10 @@ codes_dir_setup() {
 
 codes_download() {
     # If download is an rpm, also installs
-    local repo=$1
-    local qualifier=${2:-}
-    local package=${3:-}
-    local version=${4:-}
+    declare repo=$1
+    declare qualifier=${2:-}
+    declare package=${3:-}
+    declare version=${4:-}
     if [[ ! $repo =~ / ]]; then
         repo=radiasoft/$repo
     fi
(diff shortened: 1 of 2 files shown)


installers/rpm-code/codes.sh
```diff
@@ -208,44 +209,54 @@ codes_err() {
...
         "$p"
         codes_install_python_done
     fi
-    local d=${codes_dir[prefix]}/lib64
+    declare t=${module}_test
+    if codes_is_function "$t"; then
+        codes_msg "Running: $t"
+        # python adds cwd to PYTHONPATH so go to a path where
+        # cwd won't contain any python that would be added
+        mkdir -p /tmp/$(date -u +%Y%m%d.%H%M%S)
```
e-carlin, line 246: @robnagler I want to call this out. Wasn't sure about a good location.

installers/rpm-code/codes.sh
```diff
@@ -208,44 +209,54 @@ codes_err() {
...
+        cd "$_"
+        "$t"
+        cd -
```
robnagler, line 249: This is not guaranteed to work since `$t` may cd.

installers/rpm-code/codes.sh
robnagler: Good idea. You can use install_tmp_dir which will clean it up better. Also it labels the directory a bit better so you know you can delete them manually. Finally, it uses /var/tmp which usually has more room than /tmp.

robnagler (changes requested):

installers/rpm-code/codes.sh
e-carlin: 7c8960c

robnagler (approved):

PR #479, opened 2023-05-25
Fix radiasoft/sirepo#5582: use setswitchinterval instead of setcheckinterval

e-carlin:
setcheckinterval was deprecated in python 3.2 and removed in pyton 3.9


robnagler (commented):
lgtm

PR #506, opened 2023-07-14
Fix #504 build py/warpx with all dimensionalities

### LINKED ISSUE #504: Add warpx and pywarpx to the Jupyter container
Following up on the previous [issue](https://github.com/radiasoft/rsfusion/issues/2)

The installation should support all dimensions (1,2,3,RZ). Our plan is to run various configurations via rsmpi.

robnagler:
- Add codes_cmake2
- Fix codes_cmake_build to support target and simpler

### DIFF
--- a/installers/rpm-code/codes.sh
+++ b/installers/rpm-code/codes.sh
@@ -22,17 +22,26 @@ codes_cmake() {
     cmake -D CMAKE_RULE_MESSAGES:BOOL=OFF -D CMAKE_BUILD_TYPE:STRING="$t" "$@" ..
 }
 
+codes_cmake2() {
+    declare t=Release
+    if [[ ${CODES_DEBUG_FLAG:-} ]]; then
+        t=Debug
+    fi
+    cmake -S . -B build -D CMAKE_RULE_MESSAGES:BOOL=OFF -D CMAKE_BUILD_TYPE:STRING="$t" "$@"
+}
+
 codes_cmake_build() {
-    declare cmd=( cmake --build . -j$(codes_num_cores) )
+    declare target=${1:-}
+    declare cmd=( cmake --build build -j$(codes_num_cores) )
     if [[ ${CODES_DEBUG_FLAG:-} ]]; then
         cmd+=( --verbose )
     fi
-    "${cmd[@]}"
+    "${cmd[@]}" ${target:+--target $target}
 }
 
 codes_cmake_fix_lib_dir() {
     # otherwise uses ~/.local/lib64
-    find . -name CMakeLists.txt -print0 | xargs -0 perl -pi -e '/include\(GNUInstallDirs/ && ($_ .= q{
+    find . \( -name CMakeLists.txt -o -name \*.cmake \) -print0 | xargs -0 perl -pi -e '/include\(GNUInstallDirs/ && ($_ .= q{
 set(CMAKE_INSTALL_LIBDIR "lib" CACHE PATH "Library installation directory." FORCE)
 GNUInstallDirs_get_absolute_install_dir(CMAKE_INSTALL_FULL_LIBDIR CMAKE_INSTALL_LIBDIR)
 })'
--- a/installers/rpm-code/codes/warpx.sh
+++ b/installers/rpm-code/codes/warpx.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+declare _warpx_src_d
+
+warpx_main() {
+    # Uses amrex, but it seems to need to build it on its own...
+    codes_dependencies common
+    codes_download https://github.com/ECP-WarpX/WarpX.git
+    _warpx_src_d=$PWD
+    codes_cmake_fix_lib_dir
+    codes_cmake2 \
+        -D CMAKE_INSTALL_PREFIX="${codes_dir[prefix]}" \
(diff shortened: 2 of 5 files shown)


installers/rpm-code/codes/warpx.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+declare _warpx_src_d
+
+warpx_main() {
+    # Uses amrex, but it seems to need to build it on its own...
```
e-carlin, line 6: I think we could get this working by setting `WarpX_amrex_internal=OFF`. We'd need to install a different version of amrex (warpx expects 23.06 and we currently install 22.09), and we'd have to figure out why AMREXConfig.cmake isn't currently installed (maybe because we build amrex with configure instead of cmake?).  There may be other issues but that was as far as I made it.

Unpacked rscode-amrex is 388mb. I assume this adds a second copy of roughly the same size. Not sure if that meets the threshold for worth doing the work.

installers/rpm-code/codes.sh
```diff
@@ -22,17 +22,26 @@ codes_cmake() {
     cmake -D CMAKE_RULE_MESSAGES:BOOL=OFF -D CMAKE_BUILD_TYPE:STRING="$t" "$@" ..
 }
 
+codes_cmake2() {
+    declare t=Release
+    if [[ ${CODES_DEBUG_FLAG:-} ]]; then
+        t=Debug
+    fi
+    cmake -S . -B build -D CMAKE_RULE_MESSAGES:BOOL=OFF -D CMAKE_BUILD_TYPE:STRING="$t" "$@"
```
e-carlin, line 30: nice find

e-carlin (approved):

installers/rpm-code/codes.sh
robnagler: The distinction with `codes_cmake2` is that it doesn't `cd build`, which is why I made it a separate function.

installers/rpm-code/codes/warpx.sh
robnagler: I tried WarpX_amrex_internal=OFF, and it doesn't find the libraries. I then tried setting amrex_dir (or whatever they wanted), but that didn't work either. I thought about creating an issue, but it's too coupled I expect (especially with dimensionalities).

> Unpacked rscode-amrex is 388mb

That's mostly symbols. The point is moot, because amrex is linked statically inside libwarpx.1d.so, libwarpx.2d.so, etc. and again in bin/warpx.1d.MPI.OMP.DP.PDP.OPMD.QED, bin/warpx.2d.MPI.OMP.DP.PDP.OPMD.QED, etc. Each of these are only 15MB so it doesn't matter.

The real cost is compile time. They have to compile 4 versions (each of the dimensionalities) of amrex on top of warpx. This takes a lot of time, but since it works out of the box, it didn't take long to build.

e-carlin: FWIW - leaving these notes mostly for myself. I was curious about this so I played around with it some this weekend. I got the warpx cmake build to work (found amrex with all required features). But, kept running into `/usr/bin/ld: lib/libbuildInfoshared_1d.a(AMReX_buildInfo.cpp.o): relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC` when compiling warpx. I tried manually passing the flag and digging into some of the make code. In order to get AmrexConfig.cmake I found it easiest to switch to building amrex with cmake. But, the amrex docs say the developers mostly use configure/make and looking at the code I don't think they properly handle -fPIC in cmake. In the end I gave up. I don't think the compilation speed would be sped up much because we'd have to compile amrex with 3 dims which slows things down and then again for warpx. Also, we'd have to pass numerous flags to amrex copied from the warpx build config which is fragile. 

Working on this highlighted my desire to create a better dev env when working on these codes. What would work for my workflow is when an error is encountered to leave the docker container running. Drop the user into it with a shell that has all of the codes.sh funcs/vars set so the developer can start playing with things. I prefer docker because then I leave my vm clean for actually installing the code and testing it once the compilation is successful. Another thing would be to cache the codes_download somewhere. This is a minor speedup but amrex/warpx but for codes like mantid would be meaningful. Another idea along this line is to create a mounted volume for the whole src dir. That way all of that state wouldn't have to be recreated between containers. Finally, creating some tooling around populating the ~/src/yum on vms. I frequently run `dnf download ...` then `creatrepo ...` to update the local repo instead of building depdencies manually which is slow. A script that takes in the name of the code you want to do this for or all if none is specified would work well. You could even do this automatically when a code isn't found.

robnagler: I don't understand why `-fPIC` is so complicated, and why it isn't the default. The static libraries (`.a`) should not be built except for certain codes which need to be a single binary (opal, for example). Anyway, thanks for looking into it. I'm sure it was enlightening...

Good thoughts on development. I had commented out `install_clean_cmds`, but your idea is better. We need for the container to sleep forever, and then issue a comment: run `docker exec bla_bla`. I'll create an issue.

I don't know that we want to cache the compile directory. We could setup a ccache, which would help, and that could be mounted. The concern I have is there is a lot of implicit coupling not picked up by CMake/Make with caching.

PR #521, opened 2023-08-28
Fix #519 patch mpicxx & mpi.h and remove mpicxx.h

### LINKED ISSUE #519: Patch bin/mpicxx to not link with -lmpicxx
Support https://github.com/radiasoft/download/issues/518 & https://github.com/radiasoft/sirepo/issues/6220. 

See https://github.com/radiasoft/sirepo/issues/6220#issuecomment-1696135090

robnagler:
- Remove references to C++ bindings
- Leave libmpicxx.so for already compiled code
- openmc/moab: Fix 'no suitable method found' in compile
  `pymoab/core.pyx:1509:48: no suitable method found`
   Numpy change "unsigned long is not np.uint64_t"

### DIFF
--- a/installers/rpm-code/codes/openmc.sh
+++ b/installers/rpm-code/codes/openmc.sh
@@ -2,7 +2,7 @@
 
 openmc_dagmc() {
     local p="$PWD"
-    codes_download svalinn/DAGMC "$version"
+    codes_download svalinn/DAGMC develop
     codes_cmake \
         -D BUILD_STATIC_LIBS=OFF \
         -D BUILD_TALLY=ON \
@@ -17,14 +17,14 @@ openmc_main() {
     codes_yum_dependencies eigen3-devel
     codes_dependencies common
     openmc_moab
-    version=develop
     openmc_dagmc
     openmc_openmc
 }
 
 openmc_moab() {
     local p="$PWD"
-    codes_download https://bitbucket.org/fathomteam/moab.git Version5.1.0
+    # 20230827 fixes pymoab/core.pyx:1509:48: no suitable method found
+    codes_download https://bitbucket.org/fathomteam/moab.git bfccfc78e6cb3ddc02c39be437a64696bf126d86
     codes_cmake_fix_lib_dir
     # This cmake module uses python-config which doesn't work with venv
     # https://mail.python.org/archives/list/python-ideas@python.org/thread/QTCPOM5YBOKCWWNPDP7Z4QL2K6OWGSHL/
@@ -44,11 +44,10 @@ openmc_moab() {
 }
 
 openmc_openmc() {
-    codes_download openmc-dev/openmc "$version"
+    codes_download openmc-dev/openmc develop
     codes_cmake_fix_lib_dir
     CXX=mpicxx codes_cmake \
         -D CMAKE_INSTALL_PREFIX="${codes_dir[prefix]}" \
-        -D ENABLE_PYMOAB=ON \
         -D HDF5_PREFER_PARALLEL=on \
         -D OPENMC_USE_DAGMC=on \
         -D OPENMC_USE_MPI=on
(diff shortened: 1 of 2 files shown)


installers/rpm-code/codes/openmc.sh
```diff
@@ -44,11 +44,10 @@ openmc_moab() {
 }
 
 openmc_openmc() {
-    codes_download openmc-dev/openmc "$version"
+    codes_download openmc-dev/openmc develop
     codes_cmake_fix_lib_dir
     CXX=mpicxx codes_cmake \
         -D CMAKE_INSTALL_PREFIX="${codes_dir[prefix]}" \
-        -D ENABLE_PYMOAB=ON \
```
e-carlin, line 51: Why was this removed (and turned off)?

installers/rpm-code/codes/openmc.sh
```diff
@@ -17,14 +17,14 @@ openmc_main() {
...
     openmc_moab
-    version=develop
     openmc_dagmc
     openmc_openmc
 }
 
 openmc_moab() {
     local p="$PWD"
-    codes_download https://bitbucket.org/fathomteam/moab.git Version5.1.0
+    # 20230827 fixes pymoab/core.pyx:1509:48: no suitable method found
```
e-carlin, line 26: Why include the date? Git blame (or vc-region-history in emacs) will tell us that. 

I've waffled on this many times but I think it might be unnecessary to include the reason for pinning too. I think it may be better to write a very detailed commit message with an abundance of information rather than a terse comment.

e-carlin (changes requested):

installers/rpm-code/codes/openmc.sh
robnagler: > Why include the date? Git blame (or vc-region-history in emacs) will tell us that.

The date is that of the commit in the download repo. This can be hard to track down (as we found with warp). It seems useful to me since it's just a sha. 

> I've waffled on this many times but I think it might be unnecessary to include the reason for pinning too. I think it may be better to write a very detailed commit message with an abundance of information rather than a terse comment.

Git blame doesn't always work well, because text gets edited over time and moved around and the exact commit message is hard to fine. I find for pinned commits, giving some reasons is useful. Given that we maintain a lot of this little build files, it's easier to put the "footnotes" inline if it is a pinned version for a every specific issues (as opposed to a package lock). I've just been finding that we end up searching around unnecessarily when we pin a version. This has happened a few times lately. Build files are different than code. A pinned version is like another if statement, and it is better to document it. Could be documented as a function, but that's awkward for the execution environment.

e-carlin: > The date is that of the commit in the download repo 

What does that mean? The date we squashed and merged to master? If so, will the date be wrong (we are going to squash and merge today 8/29)? I don't know how github implements a sqaush merge but if they do a rebase then I believe the commit date changes to the day the rebase was performed. I believe then it leaves the author date the same. But I don't know what it picks for the author date when squashing. 

I feel fairly strongly that dates are better left to git. I use git extensively when doing work in the downloads repo and adding some other date I think will lead to confusion. For example, if I move that line of code in a future change should I change the date?

e-carlin:
@robnagler you can merge this. My comments a purely stylistic and can be changed later.

installers/rpm-code/codes/openmc.sh
robnagler: > What does that mean?

It means the date in https://bitbucket.org/fathomteam/moab.git for bfccfc78e6cb3ddc02c39be437a64696bf126d86. If that commit gets deleted or incorrectly recorded (as it did with warp somehow), we have no way of knowing when something happened. This way we know the intention.

e-carlin: Oh geez, that makes sense:)

PR #528, opened 2023-09-13
Fix #323 rpm-code/dnf-download.sh

### LINKED ISSUE #323: get all beamsim-codes rpm from depot
Add a feature to dev-build-all.sh (or maybe a new script) that downloads the latest version of each rpm from depot.

I've been trying to build all beamsim codes and have encountered errors and it is a time consuming process. Really I just want all of the rpms on my machine to I can build container-beamsim locally.

robnagler:
Most of the way there for #323

### DIFF
--- a/installers/rpm-code/dev-build.sh
+++ b/installers/rpm-code/dev-build.sh
@@ -14,9 +14,9 @@ fi
 if [[ $1 == common ]]; then
     cd ~/src/radiasoft/container-rpm-code
     if [[ ! $(docker images | grep radiasoft/fedora) ]]; then
-        cd container-fedora
-        # in case set by dev-env.sh, because server isn't running yet
+        cd ../container-fedora
         radia_run container-build
+        cd -
     fi
     radia_run container-build
 fi
--- a/installers/rpm-code/dnf-download.sh
+++ b/installers/rpm-code/dnf-download.sh
@@ -1,35 +1,19 @@
 #!/bin/bash
+#
+# Download rscode-*.rpm from system dnf repo
+#
+set -e
+declare code=$1
 set +euo pipefail
 source ~/.bashrc
 set -euo pipefail
 source ./dev-env.sh
-declare f
-sudo dnf makecache
-declare d=dnf-download-tmp
-rm -rf "$d"
-mkdir "$d"
-declare common=0
-for f in "$@"; do
-    dnf download --downloaddir "$d" "rscode-$f"
-    if [[ $1 == common ]]; then
-        common=1
-        cd ~/src/radiasoft/container-rpm-code
-        if [[ ! $(docker images | grep radiasoft/fedora) ]]; then
-            cd container-fedora
-            # in case set by dev-env.sh, because server isn't running yet
-            radia_run container-build
-        fi
-        radia_run container-build
-    fi
-done
-mv "$d"/*.rpm "$rpm_code_install_dir"
+dnf download --downloaddir "$rpm_code_install_dir" "rscode-$code" || true
 createrepo -q --update "$rpm_code_install_dir"
-if (( common )); then
+if [[ $code == common ]]; then
     cd ~/src/radiasoft/container-rpm-code
     if [[ ! $(docker images | grep radiasoft/fedora) ]]; then
-        cd container-fedora
-        # in case set by dev-env.sh, because server isn't running yet
-        radia_run container-build
+        docker pull radiasoft/fedora
     fi
-    radia_run container-build
+    install_server= radia_run container-build
 fi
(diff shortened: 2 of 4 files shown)


installers/rpm-code/dnf-download.sh
```diff
@@ -0,0 +1,35 @@
...
+        cd ~/src/radiasoft/container-rpm-code
+        if [[ ! $(docker images | grep radiasoft/fedora) ]]; then
+            cd container-fedora
+            # in case set by dev-env.sh, because server isn't running yet
+            radia_run container-build
+        fi
+        radia_run container-build
+    fi
+done
+mv "$d"/*.rpm "$rpm_code_install_dir"
```
rorour, line 25: What is the purpose of downloading to dnf-download-tmp instead of $rpm_code_install_dir?

installers/rpm-code/dnf-download.sh
```diff
@@ -0,0 +1,35 @@
...
+sudo dnf makecache
+declare d=dnf-download-tmp
+rm -rf "$d"
+mkdir "$d"
+declare common=0
+for f in "$@"; do
+    dnf download --downloaddir "$d" "rscode-$f"
+    if [[ $1 == common ]]; then
+        common=1
+        cd ~/src/radiasoft/container-rpm-code
```
rorour, line 16: Can you explain why the contents of this block (L16-22) would need to be run multiple times in this script? It's also duplicated below, so maybe should be made into a function

rorour (commented):
I don't see a simple way to download all beamsim codes at once, which seemed to be what @e-carlin was requesting in the issue

installers/rpm-code/dnf-download.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+set +euo pipefail
+source ~/.bashrc
```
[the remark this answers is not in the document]
robnagler: radia_run comes from ~/.bashrc. It's a function.

installers/rpm-code/dnf-download.sh
```diff
@@ -0,0 +1,35 @@
...
+set -euo pipefail
+source ./dev-env.sh
+declare f
+sudo dnf makecache
```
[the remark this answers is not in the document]
robnagler: I was trying to avoid this:

  [SKIPPED] rscode-common-20230829.173819-1.x86_64.rpm: Already downloaded

However, it kind of makes sense in this case to not force a download every time. This allows a refresh so I changed the code in 2ac0988.

installers/rpm-code/dnf-download.sh
```diff
@@ -0,0 +1,35 @@
...
+declare d=dnf-download-tmp
+rm -rf "$d"
+mkdir "$d"
+declare common=0
+for f in "$@"; do
+    dnf download --downloaddir "$d" "rscode-$f"
+    if [[ $1 == common ]]; then
+        common=1
```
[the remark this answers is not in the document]
robnagler: This was confused, duplicated code. :(  2ac0988

installers/rpm-code/dnf-download.sh
robnagler: 2ac0988

installers/rpm-code/dnf-download.sh
robnagler: 2ac0988

robnagler:
> I don't see a simple way to download all beamsim codes at once, which seemed to be what @e-carlin was requesting in the issue

See https://github.com/radiasoft/download/pull/528#issue-1895333923, but your comment forced me to take a look, and see how easy it was to create dnf-download-all.sh which does a better job anyway with makecache at this level. I think I fully address the issue now.

installers/rpm-code/dev-build.sh
```diff
@@ -14,9 +14,9 @@ fi
 if [[ $1 == common ]]; then
     cd ~/src/radiasoft/container-rpm-code
     if [[ ! $(docker images | grep radiasoft/fedora) ]]; then
-        cd container-fedora
-        # in case set by dev-env.sh, because server isn't running yet
+        cd ../container-fedora
         radia_run container-build
+        cd -
```
rorour, line 19: Can you explain the change to this file?

rorour (approved):
Looks good

installers/rpm-code/dev-build.sh
robnagler: It is an unrelated bug I saw. The `cd container-fedora` was relative to `cd ~/src/radiasoft/container-rpm-code`, and `container-fedora` is at the same level, and alternative coding would be:

```bash
    if [[ ! $(docker images | grep radiasoft/fedora) ]]; then
        cd ~/src/radiasoft/container-fedora
        radia_run container-build
    fi
    cd ~/src/radiasoft/container-rpm-code
    radia_run container-build
```

However, that denormalizes `~/src/radiasoft/` so it's a tiny bit better to do it the other way.

PR #538, opened 2023-09-19
Fix #536: make bluesk depend on necessary rscode's to eliminate conflicts

### LINKED ISSUE #536: bluesky conflicts
```
/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/Shadow/ShadowLib.cpython-39-x86_64-linux-gnu.so conflicts between attempted installs of rscode-bluesky-20230919.195042-1.x86_64 and rscode-shadow3-20230114.185822-1.x86_64
/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/Shadow/ShadowLibExtensions.py conflicts between attempted installs of rscode-bluesky-20230919.195042-1.x86_64 and rscode-shadow3-20230114.185822-1.x86_64
/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/srwpy-4.0.0b1.dist-info/RECORD conflicts between attempted installs of rscode-bluesky-20230919.195042-1.x86_64 and rscode-srw-20230808.154627-1.x86_64
/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/srwpy/srwlib.py conflicts between attempted installs of rscode-bluesky-20230919.195042-1.x86_64 and rscode-srw-20230808.154627-1.x86_64
/home/vagrant/.pyenv/versions/3.9.15/envs/py3/lib/python3.9/site-packages/xraylib.py conflicts between attempted installs of rscode-bluesky-20230919.195042-1.x86_64 and rscode-xraylib-20230114.185403-1.x86_64
```

e-carlin:


installers/rpm-code/codes/bluesky.sh
```diff
@@ -2,7 +2,7 @@
 
 bluesky_main() {
     codes_yum_dependencies mesa-libGL mongodb-org-server
-    codes_dependencies common ipykernel
+    codes_dependencies common ipykernel shadow3 srw xraylib
```
robnagler, line 5: shadow3 depends on xraylib so you technically don't need to expose this dependency here.

I don't understand what conflicts are being eliminated by this install. Does srw and shadow install something that is used by bluesky? If so, that should be moved into common. Technically rscode-bluesky doesn't depend on srw or shadow (for raydata).

e-carlin: > Does srw and shadow install something that is used by bluesky? If so, that should be moved into common.

bluesky depends on srwpy [through sirepo-bluesky](https://github.com/NSLS-II/sirepo-bluesky/blob/8378488e9f620b51433f4987011f205bcd4b0494/requirements.txt#L14). So we could move the install of srwpy into common but then what do we do with things like the patches in srw.sh? Do those get moved into common? I wasn't sure. Same goes for shadow. If we moved into common we would effectively get rid of both rsocde-srw and rscode-shadow3 which seemed like the wrong move. Instead I thought that since bluesky.sh depends on sirepo-bluesky and sirepo-bluesky depends on srw/shadow, and since we use our own flavor of srw we should just call the dependency out here. Definitely not the way we normally do things but seemed like the right approach in this case.

robnagler: nvm. I didn't read the whole file. You are correct. xraylib still could be pulled unless it is used explicitly.

PR #561, opened 2023-10-25
add --no-cache-dir to pip install to free space during ci run

rorour:
fix https://github.com/radiasoft/download/issues/562
I would like to try adding `--no-cache-dir` to see if it will fix the OSError that we are running into in a private repo.
Here is pip's documentation on the potential downsides of adding this flag. https://pip.pypa.io/en/stable/topics/caching/#disabling-caching

### DIFF
--- a/installers/python-ci/radiasoft-download.sh
+++ b/installers/python-ci/radiasoft-download.sh
@@ -50,7 +50,7 @@ python_ci_main() {
                 pip install git+https://'${GITHUB_TOKEN:+$GITHUB_TOKEN@}'github.com/radiasoft/\$x.git
             done
             pip uninstall -y '$r' >& /dev/null || true
-            pip install -e .
+            pip install --no-cache-dir -e .
             export PYKERN_PKCLI_TEST_MAX_FAILURES=1 PYKERN_PKCLI_TEST_RESTARTABLE=1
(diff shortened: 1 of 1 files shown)


installers/python-ci/radiasoft-download.sh
```diff
@@ -50,7 +50,7 @@ python_ci_main() {
                 pip install git+https://'${GITHUB_TOKEN:+$GITHUB_TOKEN@}'github.com/radiasoft/\$x.git
             done
             pip uninstall -y '$r' >& /dev/null || true
-            pip install -e .
+            pip install --no-cache-dir -e .
```
e-carlin, line 53: Leave a comment saying why we have this. It will be in the git log but I think this is so specific and non-standard (for us) that it is good to have a comment too. Something like:
```
# GitHub CI runners are limited on disk space so use --no-cache-dir to limit disk usage as much as
# possible. See git.radiasoft.org/downloads/issues/XXX
```

e-carlin (approved):
I think this is a good approach. Add the comment and you can merge.

installers/python-ci/radiasoft-download.sh
rorour: 90d140f5a24c095c86312cd89cff22be1f11900c

PR #564, opened 2023-10-25
fix #563: use different image for rslearn

### LINKED ISSUE #563: use different image for rslearn in python-ci installer
To use less space during tests

rorour:


installers/python-ci/radiasoft-download.sh
```diff
@@ -17,6 +17,9 @@ python_ci_main() {
             i=radiasoft/python3
             p+=( pykern )
             ;;
+        rslearn)
+            i=radiasoft/python3
```
e-carlin, line 21: I think this can be added to the case above (ex `rsaccounting|rsconf|rslearn`) because rslearn will need pykern installed to work.

e-carlin (changes requested):

installers/python-ci/radiasoft-download.sh
rorour: thanks for catching that. 5e8464c58272fad2dae3202f50a7e1e9e9d05a28

e-carlin (approved):

PR #573, opened 2023-11-15
Fix #572: update flash to 4.7.1

### LINKED ISSUE #572: flash-tarball update version
Need to run tests to ensure that the version is compatible with various packages we are using.

e-carlin:


e-carlin:
I've put the latest flash source on the build machine. We'll just need to build once this is merged.

installers/flash-tarball/radiasoft-download.sh
```diff
@@ -62,36 +62,21 @@ flash_tarball_patch_and_update_tgz() {
...
-     parseCmd.init()
-     # setup.py is FLASH_HOME/bin/setup.py
-EOF
     patch --quiet tools/python/flmake/setup.py <<'EOF'
 @@ -32,7 +32,7 @@
      global RUNTIME_FILES
- 
+
      # this is not a threaded application
 -    sys.setcheckinterval(10000)
```
robnagler, line 70: Are the two setup.py's different?

robnagler (approved):
Seems fine to me, just one question.

installers/flash-tarball/radiasoft-download.sh
e-carlin: Yes, for whatever reason tools/python/flmake/setup.py has not been fixed.

PR #587, opened 2023-12-15
Fix #586 radia.sh switch mpi_cxx to mpi

### LINKED ISSUE #586: radia links to libmpicxx but shouldn't
https://github.com/radiasoft/download/issues/519 is not sufficient since Radia setup.py explictly lists mpicxx.

robnagler:
NERSC has deleted the deprecated MPI C++ interfaces so libmpicxx is not there. Link with libmpi instead.


e-carlin (commented):
LGTM

PR #589, opened 2023-12-19
Fix #588: Create a minimal working example for vm_devbox

### LINKED ISSUE #588: vagrant-dev on centos7
Create installer for a vagrant-sirepo-dev vm on centos7

e-carlin:
Not sure that this will be the final configuration. But, good enough for starting.

- Make setting private_network optional: We are going to connect over ssh with port forwarding so we won't need a private network to talk with the host
- Use box=generic/fedora$install_version_fedora: I get a 404 when trying to access the cloud-base box for all providers.


e-carlin:
@rorour merging to keep moving but please review.

rorour (commented):
looks good to me

PR #595, opened 2024-01-03
Fix #594: create installer for beamsim jupyter scientific packages

### LINKED ISSUE #594: create beamsim-jupyter installer
We need to install the same but different deps when installing for beamsim-jupyter vs jupyter-nvidia (ex tensorflow for cpu or tensorflow for gpu). So, create an installer that both can call that breaks out the different bits.

e-carlin:
Dependencies like jupyterhub are still installed in container-beamsim-jupyter/container-conf/buid.sh. But, all scientific packages (ex tensorflow) are now installed here.

This makes it easy to see everything in one place and make decisisions to handle gpu vs cpu.

In addition, tensorflow was updated. We now just install tensorflow and have pip handle the dependencies. The previous behavior of installing dependencies ourselves was to get around a bug we now hope is resolved.

### DIFF
--- a/installers/beamsim-jupyter/radiasoft-download.sh
+++ b/installers/beamsim-jupyter/radiasoft-download.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+#
+# Install dependencies common to beamsim-jupyter and container-jupyter-nvidia but
+# install the correct version for gpu vs not.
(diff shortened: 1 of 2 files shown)


e-carlin:
Need to add one more thing.

PR #597, opened 2024-01-05
Fix #594: create installer for beamsim jupyter scientific packages

### LINKED ISSUE #594: create beamsim-jupyter installer

e-carlin:
Dependencies like jupyterhub are still installed in container-beamsim-jupyter/container-conf/buid.sh. But, all scientific packages (ex tensorflow) are now installed here.

This makes it easy to see everything in one place and make decisisions to handle gpu vs cpu.

In addition, tensorflow was updated. We now just install tensorflow and have pip handle the dependencies. The previous behavior of installing dependencies ourselves was to get around a bug we now hope is resolved.

Finally, bump hypre version to get around a compile error with newer cpp compilers https://github.com/hypre-space/hypre/pull/731.


e-carlin:
@robnagler if possible please review today. These changes need to run through a full release cycle before we can update gpu jupyter.

installers/beamsim-jupyter/radiasoft-download.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+#
+# Install dependencies common to beamsim-jupyter and container-jupyter-nvidia but
+# install the correct version for gpu vs not.
+#
+
+beamsim_jupyter_main() {
```
robnagler, line 7: alphabetic sort order

installers/beamsim-jupyter/radiasoft-download.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
robnagler, line 1: The installer name is confusing because it neither installs beamsim codes nor jupyter. This is a specialized installer, which could be documented as "pip-install-ml". We have installers like "sirepo", "jupyter", etc. that install the docker images and run them. One might think that this installer as named would download and run the beamsim-jupyter docker image.

installers/beamsim-jupyter/radiasoft-download.sh
```diff
@@ -0,0 +1,79 @@
...
+#
+# Install dependencies common to beamsim-jupyter and container-jupyter-nvidia but
+# install the correct version for gpu vs not.
+#
+
+beamsim_jupyter_main() {
+    declare gpu_only=${1:-}
```
robnagler, line 8: For flags like this, it's annoying. I think instead you should have something like beamsim_jupyter_gpu=1. This way the code is documented. It's not "gpu_only" because the other is "cpu_only". It's gpu or not gpu, so you don't need the only. Alternatively, you could have:

```sh
declare mode=$1
case $mode in
gpu)
cpu)
*)
install_err
```

There's no need to take shortcuts here especially with installers, since there's only a little bit of code needed and it documents the intention of something we rarely visit.

robnagler (changes requested):

installers/beamsim-jupyter/radiasoft-download.sh
e-carlin: `gpu_only` is what you recommended in the past https://github.com/radiasoft/download/pull/187#discussion_r770979574. This case seemed similar to me so rather than invent a new concept I thought it made sense to stick with what we already have.  "shortcut" never crossed my mind...

installers/beamsim-jupyter/radiasoft-download.sh
e-carlin: I don't like `pip-install-ml`. I think both `pip` and `ml` set unnecessary and incorrect (ex seaborn is not strictly for ml) expectations. How about `jupyter-scientific-packages`?

installers/beamsim-jupyter/radiasoft-download.sh
e-carlin: 1a84e64

installers/beamsim-jupyter/radiasoft-download.sh
robnagler: > elegant_install_bin "$gpu_only" (which, since internal, can just be a 1 or empty).

The emphasis is on "since internal". This is all within one file. The parameter in this case is being passed from one installer to another.

installers/beamsim-jupyter/radiasoft-download.sh
robnagler: - These have nothing to do with Jupyter. They can be used outside Jupyter just fine.
- The install is specifically for python related packages, and they do use pip (vs conda) so I think pip describes what this is doing specifically, and differentiates from other types of installers.
- Everything we do with beamsim is scientific :)
- Packages is superfluous

How about `py-ml-and-coupled`? The problem is that there's a coupling between the ml packages and seaborn, etc.

e-carlin: > The install is specifically for python related packages, and they do use pip (vs conda) so I think pip describes what this is doing specifically, and differentiates from other types of installers.

I guess that is a distinction I didn't know we were enforcing. I'm not sure why it is useful to say "we can only install python here".

> Everything we do with beamsim is scientific :)

Yes, not a great name. My hope was to distinguish between deps like jupyterlab which are more for the "service" and less for the "science".

> Packages is superfluous

Agreed.

> How about py-ml-and-coupled? The problem is that there's a coupling between the ml packages and seaborn, etc.

I'm hung up on ml being in the name. I don't understand why the packages in this installer __have__ to have something to do with ml?

installers/beamsim-jupyter/radiasoft-download.sh
e-carlin: `gpu-only` is set by another installer https://github.com/radiasoft/container-jupyter-nvidia/blob/522bc38178df77576b493e783bfece1d50ae0099/container-conf/build.sh#L36

robnagler: gpu-only is not correct if it can't be used on a cpu. Is that the case here?

installers/beamsim-jupyter/radiasoft-download.sh
robnagler: I think it is replacing ml.sh almost entirely, and frankly, it seems like it should replace it all.

>    install_pip_install h5ImageGenerator \
>        scikit-learn sympy \
>        tensorflow~=2.15.0

I think this is confusing in ml.sh. I think we should build two rpms rscode-ml-cpu and rscode-ml-gpu. perhaps they could provide "rscode-ml" so the dependencies make sense, and gpu-jupyter docker image should replace rscode-ml-cpu with rscode-ml-gpu.

I think we should installing over an rscode rpm. Indeed, it might make sense for beamsim-jupyter to be built with other rpms, and not have installs of its own except for configuration files that vary from sirepo. Even those could be rpms.

If you need to release this, do that. However, I think this has been a mess for a while, and we need to figure out how to make it clear/clean.

installers/beamsim-jupyter/radiasoft-download.sh
e-carlin: I hate to say this but that double negative is making your sentence hard to understand. Can you say it a different way?

robnagler: In this case is "gpu-only" meaning the other case is "gpu and cpu" or is gpu-only meaning "gpu instead of cpu"? I think the flag gpu-only may be incorrect, IOW.

installers/beamsim-jupyter/radiasoft-download.sh
e-carlin: After discussion: We are going to move all ml deps into this installer. Then ml.sh will call this installer (gpu vs non-gpu). Packages like seaborn will go back to container-beamsim-jupyter

e-carlin:
@robnagler please review again.

installers/ml-python/radiasoft-download.sh
```diff
@@ -0,0 +1,61 @@
...
+# We need to pin versions because package versions need to be kept in
+# sync with gpu software versions (cuda).
+ml_python_gpu() {
+    install_pip_install --index-url https://download.pytorch.org/whl/cu121 \
+        "${_ml_python_torch_base[@]}"
+    install_pip_install --find-links https://data.pyg.org/whl/torch-2.1.0+cu121.html \
+        "${_ml_python_torch_extras[@]}"
+    # Tensorflow, python, cuda, compatability matrix
+    # https://www.tensorflow.org/install/source#gpu
+    install_pip_install tensorflow[and-cuda]~=_ml_python_tensorflow_version
```
robnagler, line 60: This is missing a $ and quotes.

robnagler (changes requested):

installers/ml-python/radiasoft-download.sh
e-carlin: 86c8dbf

robnagler (approved):

PR #599, opened 2024-01-16
Fix #598: support vm_devbox

### LINKED ISSUE #598: create single arg for vagrant_dev_vm_devbox and allow setting of hostname
https://github.com/radiasoft/rsconf/pull/432#discussion_r1450809086

e-carlin:
- Allow setting of single variable vagrant_dev_vm_devbox that sets correct defaults for other variables
- support non v or radia.run hostnames

### DIFF
--- a/installers/vagrant-dev/radiasoft-download.sh
+++ b/installers/vagrant-dev/radiasoft-download.sh
@@ -75,6 +75,10 @@ vagrant_dev_ignore_git_dir_ownership() {
 
 vagrant_dev_ip() {
     declare host=$1
+    if [[ ! ${vagrant_dev_provision_eth1:-} ]]; then
+        echo -n
+        return
+    fi
     declare i=$(dig +short "$host" 2>/dev/null || true)
     if [[ $i ]]; then
         echo -n "$i"
@@ -122,6 +126,9 @@ vagrant_dev_main() {
             v|v[1-9]|*.radia.run)
                 host=$a
                 ;;
+            [a-z]*.*.*)
+                host=$a
+                ;;
             *)
                 install_err "invalid arg=$a
 expects: fedora|centos[/<version>], <ip address>, update, v[1-9].radia.run"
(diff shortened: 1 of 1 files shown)


installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -140,16 +147,23 @@ expects: fedora|centos[/<version>], <ip address>, update, v[1-9].radia.run"
...
     fi
-    vagrant_dev_private_net=${vagrant_dev_private_net-1}
-    # vbguest is pretty broken so we are defaulting to off for now
-    vagrant_dev_no_vbguest=${vagrant_dev_no_vbguest-1}
+    if [[ ${vagrant_dev_vm_devbox:+1} ]]; then
+        vagrant_dev_no_mounts=${vagrant_dev_no_mounts-1}
+        vagrant_dev_no_nfs_src=${vagrant_dev_no_nfs_src-1}
+        vagrant_dev_no_vbguest=${vagrant_dev_no_vbguest-1}
+        vagrant_dev_private_net=${vagrant_dev_private_net-}
+        vagrant_dev_provision_eth1=${vagrant_dev_provision_eth1-}
```
robnagler, line 155: `vagrant_dev_provision_eth1=${vagrant_dev_provision_eth1-1}`

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -140,16 +147,23 @@ expects: fedora|centos[/<version>], <ip address>, update, v[1-9].radia.run"
...
+    fi
     if [[ ${vagrant_dev_barebones:+1} ]]; then
         # allow individual overrides
         vagrant_dev_no_dev_env=${vagrant_dev_no_dev_env-1}
         vagrant_dev_no_mounts=${vagrant_dev_no_mounts-1}
         vagrant_dev_no_nfs_src=${vagrant_dev_no_nfs_src-1}
         vagrant_dev_no_vbguest=${vagrant_dev_no_vbguest-1}
     fi
+    vagrant_dev_private_net=${vagrant_dev_private_net-1}
+    # vbguest is pretty broken so we are defaulting to off for now
```
robnagler, line 165: This is defaulting to on.

robnagler (changes requested):

installers/vagrant-dev/radiasoft-download.sh
e-carlin: Why? We aren't handling the the rsconf case right now so I don't think eth1 needs to be properly setup. If we are going to turn it on then we are going to have to figure out ip addresses for these machines.

installers/vagrant-dev/radiasoft-download.sh
e-carlin: There is a "no" in the variable name. So it is setting no vbguest to true which means vbguest as a whole is off.

installers/vagrant-dev/radiasoft-download.sh
robnagler: Ah, didn't realize. Ignore.

installers/vagrant-dev/radiasoft-download.sh
robnagler: Love those double negatives... It's really hard to get the names right without "no" when they are optional...

robnagler (approved):

PR #602, opened 2024-01-17
Fix #601: create rscode-julia

### LINKED ISSUE #601: create rscode-julia
https://github.com/radiasoft/container-beamsim-jupyter/issues/95

e-carlin:


installers/rpm-code/codes/julia.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+julia_main() {
+    codes_dependencies common
+    declare v=julia-1.9.3
+    declare d=${codes_dir[prefix]}/$v
+    install -d -m 755 "$d"
+    codes_curl https://julialang-s3.julialang.org/bin/linux/x64/1.9/"$v"-linux-x86_64.tar.gz \
```
robnagler, line 8: probably should make sure permissions are right in the tar: chmod -R a+rX

robnagler (changes requested):

installers/rpm-code/codes/julia.sh
e-carlin: I didn't know about `X`. That's neat.

781001e

robnagler (approved):

PR #603, opened 2024-01-18
Fix #600: Move dev-setup.sh into dev-env.sh

### LINKED ISSUE #600: Errors building common for the first time
Trying to run `bash dev-build.sh common` from a clean vm. I got the following errors in order:
- `install: cannot create regular file '/home/vagrant/src/yum/fedora/36/x86_64/dev/rscode-common-20240117.180905-1.x86_64.rpm': No such file or directory`. Fixed with `mkdir -p /home/vagrant/src/yum/fedora/36/x86_64/dev/`
- `/var/tmp/radia-run-31919-30613/20240117182827-radiasoft-download.sh: line 116: createrepo: command not found`. Fixed with `yum install -y createrepo`
- `dev-build.sh: line 15: cd: /home/vagrant/src/radiasoft/container-rpm-code: No such file or directory`. Fixed with `gcl container-rpm-code`.
- `curl -L -s -S http://10.10.10.20:2916/radiasoft/containers/radiasoft-download.sh
radiasoft-download.sh: no #! at start of file: /var/tmp/radia-run-68756-31957/20240117203424-radiasoft-download.sh`. Fixed with `gcl containers`
- `#7 3.371 Status code: 404 for http://10.10.10.20:2916/yum/fedora/36/x86_64/dev/radiasoft.repo (IP: 10.10.10.20)`. 

After encountering those I realized that `rpm-code/dev-setup.sh` will fix them. That is called by `rpm-code/dev-server.sh`. I ran `etc/dev-server.sh`. I think we can ditch `rpm-code/dev-server.sh` and have `etc/dev-server.sh` call `rpm-code/dev-server.sh`.

e-carlin:
dev-env.sh is called in more places (ex dev-server.sh and dev-build.sh) so we have a better chance of making sure the system will be properly setup regardless of what the developer is doing.

### DIFF
--- a/installers/rpm-code/dev-env.sh
+++ b/installers/rpm-code/dev-env.sh
@@ -26,3 +26,7 @@ export install_proprietary_key=proprietary_code
 # for convenience to test rpm-perl, not used here
 export rpm_perl_install_dir=$(_root)/radiasoft/rsconf/rpm
 unset _root
+if [[ ! -e $radiasoft_repo_file ]]; then
+    echo 'setting up one time'
+    bash dev-setup.sh
+fi
--- a/installers/rpm-code/dev-server.sh
+++ b/installers/rpm-code/dev-server.sh
@@ -1,9 +1,5 @@
 #!/bin/bash
 set -euo pipefail
 source ./dev-env.sh
-if [[ ! -e $radiasoft_repo_file ]]; then
-    echo 'setting up one time'
-    bash dev-setup.sh
-fi
 cd ~/src
 PYENV_VERSION=py3 exec pyenv exec python -m http.server "$dev_port"


installers/rpm-code/dev-env.sh
```diff
@@ -26,3 +26,7 @@ export install_proprietary_key=proprietary_code
 # for convenience to test rpm-perl, not used here
 export rpm_perl_install_dir=$(_root)/radiasoft/rsconf/rpm
 unset _root
+if [[ ! -e $radiasoft_repo_file ]]; then
```
robnagler, line 29: Seems like this should be embedded in dev-setup.sh so that the sentinel is modular.

robnagler (changes requested):

installers/rpm-code/dev-env.sh
robnagler: There is recursion with dev-env.sh now so this won't work anyway. 

I think this code could use more functions and possibly be a single file, e.g. dev-env.sh, which does all the work. It could take an argument "server". If the server isn't running, it would tell the user to run dev-env.sh server.

robnagler (changes requested):

PR #606, opened 2024-02-01
Fix #605: disable services that prevent custom ssh

### LINKED ISSUE #605: vagrant_dev_vm_devbox allow ssh connections
- disable firewalld
- disable selinux

e-carlin:


installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -375,6 +376,17 @@ vagrant_dev_provision_private_net() {
 EOF
 }
 
+vagrant_dev_stop_services() {
+    vagrant ssh <<EOF
+sudo bash -s <<EOF_BASH
```
robnagler, line 381: I don't think you need the `-s`.

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -375,6 +376,17 @@ vagrant_dev_provision_private_net() {
 EOF
 }
 
+vagrant_dev_stop_services() {
+    vagrant ssh <<EOF
```
robnagler, line 380: I think it would be better to quote the `'EOF'` unless you need interpolation. Same for EOF_BASH

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -197,6 +197,7 @@ expects: fedora|centos[/<version>], <ip address>, update, v[1-9].radia.run"
     if [[ ${vagrant_dev_no_dev_env:+1} ]]; then
         return
     fi
+    vagrant_dev_stop_services
```
robnagler, line 200: `vagrant_dev_disable_security`

PR #608, opened 2024-02-06
Fix #607: resolve pr comments

### LINKED ISSUE #607: fix pr comments
https://github.com/radiasoft/download/pull/606#pullrequestreview-1858089580

e-carlin:


installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -40,6 +40,17 @@ vagrant_dev_box_add() {
     fi
 }
 
+vagrant_dev_disable_security() {
+    vagrant ssh <<'EOF'
+sudo bash <<'EOF_BASH'
+systemctl stop firewalld || true
+systemctl disable firewalld || true
+perl -pi -e 's{(?<=^SELINUX=).*}{disabled}' /etc/selinux/config || true
```
robnagler, line 48: I don't think you want the `|| true` for the edit. If /etc/selinux/config isn't there, perl will crash. Otherwise, perl will exit 0.

robnagler (changes requested):

installers/vagrant-dev/radiasoft-download.sh
e-carlin: a3494c1

PR #695, opened 2024-07-08
Fix #649: mlopal for gpu-jupyter

### LINKED ISSUE #649: Move p4p out of epics
p4p installs a lot of stuff in python that epics IOCs don't use and potentially conflicts with.

e-carlin:


robnagler (approved):
Didn't test

PR #698, opened 2024-07-09
fix #697 add neutronics_material_maker for openmc sims

### LINKED ISSUE #697: Add neutronics_material_maker to openmc code
Add this special pip install to openmc.sh ```pip install "neutronics_material_maker[density]"```

moellep:
@e-carlin this pip install requires an odd formatted name. I'm not sure if it will work with the radiasoft installer

### DIFF
--- a/installers/rpm-code/codes/bluesky.sh
+++ b/installers/rpm-code/codes/bluesky.sh
@@ -28,7 +28,6 @@ bluesky_main() {
         photutils \
         pyCHX==4.0.10 \
         pyOlog \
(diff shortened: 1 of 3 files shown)


e-carlin:
@moellep that should work. I was trying to test but looks like openmc was updated to only [support python >= 3.10](https://github.com/openmc-dev/openmc/blob/develop/setup.py#L62) (we have 3.9).  So, I'll take it from here to find a version that works for us.

e-carlin (approved):

PR #701, opened 2024-07-18
Fix #700: vagrant-dev with alma

### LINKED ISSUE #700: vm with almalinux/9
Get a vagrant-dev vm running. Make sure redhat-docker works too.

e-carlin:
The name centos is left for backwards compat. One can also supply the name alma to the installer.

This doesn't resolve all possible issues related to CentOS in downloads but just gets the vagrant-dev installer working.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -45,6 +45,12 @@ install_args() {
     fi
 }
 
+install_assert_centos() {
+    if [[ $install_os_release_id != almalinux ]]; then
+        install_err 'only works on RHEL like Linux'
+    fi
+}
+
 install_assert_pip_version() {
     declare package=$1
     declare version=$2
@@ -204,7 +210,7 @@ install_init_vars() {
     : ${install_version_fedora:=36}
     : ${install_version_python:=3.9.15}
     : ${install_version_python_venv:=py${install_version_python%%.*}}
-    : ${install_version_centos:=7}
+    : ${install_version_centos:=9}
     install_init_vars_oci
     eval "$(install_vars_export)"
 }
@@ -509,16 +515,12 @@ install_version_fedora_lt_36() {
 
 install_yum() {
     declare args=( "$@" )
-    declare yum=yum
-    if [[ $(type -t dnf) ]]; then
-        yum=dnf
-    fi
     declare flags=( -y --color=never )
     if [[ ! $install_debug ]]; then
         flags+=( -q )
     fi
-    install_info "$yum" "${args[@]}"
-    install_sudo "$yum" "${flags[@]}" "${args[@]}"
+    install_info dnf "${args[@]}"
+    install_sudo dnf "${flags[@]}" "${args[@]}"
 }
 
 install_yum_install() {
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -236,7 +236,7 @@ install_init_vars_os_release() {
     fi
     # COMPAT: darwin doesn't support ${x,,}
     export install_os_release_id=$(uname | tr A-Z a-z)
-    if [[ $install_os_release_id == darwin ]]; then
+    if install_os_is_darwin; then
         export install_os_release_version_id=$(sw_vers -productVersion)
     else
         # Have something legal; unlikely to get here
@@ -289,13 +289,17 @@ install_os_is_almalinux() {
     [[ $install_os_release_id =~ almalinux ]]
 }
 
-install_os_is_centos() {
+install_os_is_rhel_compatible() {
     install_os_is_almalinux || [[ $install_os_release_id =~ centos ]]
 }
 
 
 install_os_is_centos_7() {
-    install_os_is_centos && $install_version_centos -eq 7
+    install_os_is_rhel_compatible && $install_version_centos -eq 7
+}
+
+install_os_is_darwin() {
+    [[ $install_os_release_id =~ darwin ]]
 }
 
 install_os_is_fedora() {
@@ -303,7 +307,7 @@ install_os_is_fedora() {
 }
 
 install_os_is_redhat() {
-    [[ $install_os_release_id =~ rhel ]] || install_os_is_fedora || install_os_is_centos
+    [[ $install_os_release_id =~ rhel ]] || install_os_is_fedora || install_os_is_rhel_compatible
 }
 
 install_pip_install() {
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -45,12 +45,6 @@ install_args() {
     fi
 }
 
-install_assert_centos() {
-    if [[ $install_os_release_id != almalinux ]]; then
-        install_err 'only works on RHEL like Linux'
-    fi
-}
-
 install_assert_pip_version() {
     declare package=$1
     declare version=$2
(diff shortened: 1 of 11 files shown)


bin/install.sh
```diff
@@ -45,6 +45,12 @@ install_args() {
     fi
 }
 
+install_assert_centos() {
+    if [[ $install_os_release_id != almalinux ]]; then
```
robnagler, line 49: This needs to be almalinux|centos. `install_assert_os_centos`

Also this should be two functions: install_os_is_centos and install_os_assert_centos. I find the latter probably is less useful than the former so could just be one function, and the clients just call install_err

bin/install.sh
```diff
@@ -204,7 +210,7 @@ install_init_vars() {
     : ${install_version_fedora:=36}
     : ${install_version_python:=3.9.15}
     : ${install_version_python_venv:=py${install_version_python%%.*}}
-    : ${install_version_centos:=7}
```
robnagler, line 207: We should have a staged release here. It should work with centos 7 for now and testers should set install_version_centos=9 in their environments. Once alma 9 is in production, we update this default.

bin/install.sh
```diff
@@ -509,16 +515,12 @@ install_version_fedora_lt_36() {
 
 install_yum() {
     declare args=( "$@" )
-    declare yum=yum
-    if [[ $(type -t dnf) ]]; then
```
robnagler, line 513: Leave backward compatible.

installers/redhat-dev/radiasoft-download.sh
```diff
@@ -3,7 +3,7 @@
 # To run: curl radia.run | bash -s redhat-dev
 #
 redhat_dev_main() {
-    if [[ ! $install_os_release_id =~ fedora|centos|rhel ]]; then
+    if [[ ! $install_os_release_id =~ fedora|centos|rhel|alma ]]; then
```
robnagler, line 6: Centralize this install_os_is_redhat.

installers/redhat-docker/radiasoft-download.sh
```diff
@@ -32,19 +32,15 @@ Then re-run this command
         perl -pi -e 's{(?<=^SELINUX=).*}{disabled}' /etc/selinux/config
         install_err 'Disabled selinux. You need to "vagrant reload", then re-run this installer'
     fi
-    if type dnf >& /dev/null; then
```
robnagler, line 35: Leave this backward compatible and adapted to rhel 9.

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -14,20 +16,17 @@ vagrant_dev_box_add() {
...
-        provider=libvirt
-    fi
     if [[ $vagrant_dev_box ]]; then
         box=$vagrant_dev_box
     elif [[ $box =~ fedora ]]; then
         if [[ $box == fedora ]]; then
             box=generic/fedora$install_version_fedora
         fi
-    elif [[ $box == centos ]]; then
+    elif [[ $box == centos || $box == alma ]]; then
```
robnagler, line 25: Make backwards compatible. Convert the argument on entry. Centralize the name mapping in a separate function.

In main we except alma* which could mean almalinux. I don't know what the right name is. This code should be abstracted to return the box name from the arg, which is an abstract entity. The parsing shouldn't be here.

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -14,20 +16,17 @@ vagrant_dev_box_add() {
     # Returns: $box
     box=$1
     declare provider=virtualbox
-    if [[ $_vagrant_dev_host_os == ubuntu ]]; then
```
robnagler, line 17: Why is this being removed?

installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -123,7 +122,7 @@ vagrant_dev_main() {
     vagrant_dev_modifiers
     for a in "$@"; do
         case $a in
-            fedora*|centos*)
+            fedora*|centos*|alma*)
```
robnagler, line 125: May want to do more syntax checking and call this $box not $os. There is code in vagrant_dev_eth1 that should be abstracted (e.g. `$(vagrant_dev_is_fedora $box)`). 

The name should be almalinux, not alma. We don't want to create our own name. There's a bug on line 27 where it is alma, but should be almalinux.

installers/redhat-base/radiasoft-download.sh
```diff
@@ -25,6 +25,10 @@ EOF
     elif [[ ! -e /etc/yum.repos.d/epel.repo ]]; then
         yum --color=never --enablerepo=extras install -y -q epel-release
     fi
+    if [[ $install_os_release_id == almalinux ]]; then
```
robnagler, line 28: Abstract.

robnagler (changes requested):
Do not remove backwards compatibility

installers/vagrant-dev/radiasoft-download.sh
e-carlin: I started down the path of removing ubuntu because it is untested for some time now and completely untested for alma. But, then didn't finish. I'll add it back in and we'll just see what happens if we ever use it.

installers/redhat-base/radiasoft-download.sh
e-carlin: d0e3a95

installers/vagrant-dev/radiasoft-download.sh
e-carlin: d0e3a95

installers/redhat-docker/radiasoft-download.sh
```diff
@@ -32,19 +32,32 @@ Then re-run this command
...
+    declare o=rhel
+    if [[ $install_os_release_id == fedora ]]; then
```
robnagler, line 36: is_fedora

installers/redhat-docker/radiasoft-download.sh
```diff
@@ -32,19 +32,32 @@ Then re-run this command
...
-        dnf -y -q install docker-ce
-    else
+    elif install_os_is_centos_7; then
         yum-config-manager \
             --add-repo https://download.docker.com/linux/centos/docker-ce.repo
         yum makecache fast
-        yum -y -q install yum-plugin-ovl
-        yum -y -q install docker-ce
+        install_yum_install yum-plugin-ovl
+    else
```
robnagler, line 55: Probably good at this point to ask is_rhel and fail in the else case.

robnagler (changes requested):

installers/redhat-docker/radiasoft-download.sh
e-carlin: 736a7be

installers/redhat-docker/radiasoft-download.sh
e-carlin: 736a7be

bin/install.sh
```diff
@@ -285,6 +285,27 @@ install_not_strict_cmd() {
     set -euo pipefail
 }
 
+install_os_is_almalinux() {
+    [[ $install_os_release_id =~ almalinux ]]
+}
+
+install_os_is_centos() {
```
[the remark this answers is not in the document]
e-carlin: 736a7be

robnagler (approved):

PR #708, opened 2024-07-31
Fix #707: Flag to set dev vs rsconf environments.

### LINKED ISSUE #707: redhat-docker support rsconf vs "dev" mode
Want to call the installer from rsconf so we need some way of deciding what context we are in https://github.com/radiasoft/rsconf/issues/518.

e-carlin:
This installer is going to be called by rsconf to setup docker on rsconf managed hosts. We only want to to do the tls, systemd, daemon.json, and systemctl steps on "dev" hosts. On rsconf managed hosts rsconf will manage the files and do the systemctl's.

### DIFF
--- a/installers/redhat-docker/radiasoft-download.sh
+++ b/installers/redhat-docker/radiasoft-download.sh
@@ -25,23 +25,13 @@ umount /var/lib/docker
 rmdir /var/lib/docker
 perl -pi -e "s{^/var/lib/docker.*}{}" /etc/fstab
 
-Then re-run this command
+Then re-run this command. This should only happen in development environments.
 '
     fi
     if selinuxenabled; then
         perl -pi -e 's{(?<=^SELINUX=).*}{disabled}' /etc/selinux/config
         install_err 'Disabled selinux. You need to "vagrant reload", then re-run this installer'
     fi
-    declare o=rhel
-    if install_os_is_fedora; then
-        dnf -y -q install dnf-plugins-core
-        o=fedora
-    fi
-    dnf -q config-manager \
-        --add-repo \
-        https://download.docker.com/linux/"$o"/docker-ce.repo
-    dnf -y -q install docker-ce
-
     if install_os_is_fedora; then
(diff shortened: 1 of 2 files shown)


installers/redhat-docker/radiasoft-download.sh
```diff
@@ -60,6 +50,9 @@ Then re-run this command
         install_err "installer does not support os=$install_os_release_id"
     fi
     install_yum_install docker-ce
+    if [[ ! ${redhat_docker_dev:-} ]]; then
```
e-carlin, line 53: @robnagler "dev" is crummy. I wasn't sure how to pull all the concepts below into one name. Thoughts?

Also, it is a bit annoying that we will have to add this flag when setting up dev machines. My thoughts for doing it this way was that it was safer. Better to have dev machines not do something they should do than rsconf managed machines (prod) do something they shouldn't do.

robnagler: install_channel?

e-carlin: I don't think that is granular enough. We could be in a "dev" rsconf setup or a "dev" setup outside of rsconf.

robnagler: What is the case you are setting redhat_docker_dev now?

e-carlin: In vm_devbox if someone wants to run with docker then they'll run `redhat_docker_dev=1 radia_run redhat-docker`.

robnagler: Maybe we could know we are inside a VM. It has to have some unique state that we can check. This seems like a feature test, for example, you could look at sudoers for vagrant (sudo -l -U vagrant | grep 'not allowed').

e-carlin: Ok, I'll do that. I'm going to close this because I need the work along with work in another branch.

PR #709, opened 2024-08-02
Fix #703: vagrant-rsconf-dev working for almalinux

### LINKED ISSUE #703: rsconf-dev support almalinux

### LINKED ISSUE #704: perl command not found in vagrant_dev_disable_security
It is early in the install so perl isn't installed yet.

### LINKED ISSUE #705: centos7 eol update mirror
Need to apply mirror change to new vms

See first step in post install:
https://github.com/radiasoft/devops/wiki/CentOS7#post-installation

### LINKED ISSUE #707: redhat-docker support rsconf vs "dev" mode

e-carlin:
This installer is going to be called by rsconf to setup docker on rsconf managed hosts. We only want to to do the tls, systemd, daemon.json, and systemctl steps on "dev" hosts. On rsconf managed hosts rsconf will manage the files and do the systemctl's.

Fix #705: centos7 is eol so need to use radiasoft depot mirror.
Fix #704: perl not yet installed so use sed
Fix #707: Flag to set dev vs rsconf environments.

### DIFF
--- a/installers/redhat-docker/radiasoft-download.sh
+++ b/installers/redhat-docker/radiasoft-download.sh
@@ -25,23 +25,13 @@ umount /var/lib/docker
 rmdir /var/lib/docker
 perl -pi -e "s{^/var/lib/docker.*}{}" /etc/fstab
 
-Then re-run this command
+Then re-run this command. This should only happen in development environments.
 '
     fi
     if selinuxenabled; then
         perl -pi -e 's{(?<=^SELINUX=).*}{disabled}' /etc/selinux/config
         install_err 'Disabled selinux. You need to "vagrant reload", then re-run this installer'
     fi
-    declare o=rhel
-    if install_os_is_fedora; then
-        dnf -y -q install dnf-plugins-core
-        o=fedora
-    fi
-    dnf -q config-manager \
-        --add-repo \
-        https://download.docker.com/linux/"$o"/docker-ce.repo
-    dnf -y -q install docker-ce
-
     if install_os_is_fedora; then
         install_yum_install dnf-plugins-core
         dnf -q config-manager \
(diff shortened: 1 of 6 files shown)


installers/redhat-docker/radiasoft-download.sh
```diff
@@ -60,6 +50,9 @@ Then re-run this command
         install_err "installer does not support os=$install_os_release_id"
     fi
     install_yum_install docker-ce
+    if [[ ! ${redhat_docker_dev:-} ]]; then
```
e-carlin, line 53: @robnagler bringing [our conversation](https://github.com/radiasoft/download/pull/708#discussion_r1699085015) back up. I don't think there is a test for the "computing context" that will work. The test we need to know is are we in rsconf or not. Any ideas?

robnagler: I think it's pretty clear that we do not want to execute `usermod -aG docker vagrant` arbitrarily. The case where we want to allow this is a virtual machine that is set up for development. The usermod gives the user root privs without a password so the test to see whether the user vagrant is in sudoers with nopassword is the right one. This will be true in vagrant VMs.

e-carlin: I don't think we want it to be true in v3/v4 vms managed by rsconf. There will be unnecessary differences between what we are testing in dev and what happens in higher envs.

For a "v" (or vm_devbox) vm then I agree.

robnagler: We can have the default for the environment variable, and then override it in this case. There are already special things we do to run rsconf in those environments.

e-carlin: 7c892d9

robnagler (approved):

PR #715, opened 2024-09-05
Fix #714 install_git_clone add GITHUB_TOKEN if appropriate

### LINKED ISSUE #714: pass github token to clone if available

robnagler:


bin/install.sh
```diff
@@ -295,7 +298,7 @@ install_os_is_rhel_compatible() {
 
 
 install_os_is_centos_7() {
-    install_os_is_rhel_compatible && $install_version_centos -eq 7
+    [[ $install_os_release_id =~ centos ]] && [[ $install_os_release_version_id == 7 ]]
```
e-carlin, line 301: Why this change? I thought the name "centos" we were going to use for alma or centos?

PR #717, opened 2024-09-09
Fix #716 added build module

### LINKED ISSUE #716: Add build to pyenv

robnagler:
Removed tox, should not be necessary at this level

### DIFF
--- a/installers/pyenv/radiasoft-download.sh
+++ b/installers/pyenv/radiasoft-download.sh
@@ -31,7 +31,7 @@ pyenv_install() {
     pyenv install "$install_version_python"
     pyenv global "$install_version_python"
     pip install --upgrade pip
-    pip install --upgrade setuptools tox
+    pip install --upgrade build setuptools
     pyenv virtualenv "$install_version_python" "$install_version_python_venv"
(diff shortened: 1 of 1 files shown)


e-carlin (approved):
LGTM. But, please provide a detailed description of the why in the commit or issue. It is totally unclear to me which dependencies we need when and why. 

Some questions to answer:
- why is tox being removed?
- why is build being added? 
- when is build used?
- why is build being installed globally here instead of in relevant pryproject.toml files? 
- Same for setuptools?

robnagler:
> Some questions to answer:
> 
> * why is tox being removed?
> * why is build being added?
> * when is build used?
> * why is build being installed globally here instead of in relevant pryproject.toml files?
> * Same for setuptools?

https://packaging.python.org/en/latest/guides/modernize-setup-py-project/ explains some. Tox was needed when it was used, but it was only used in development. Since pip install didn't work for any of our modules before pyproject.toml, we didn't run python setup.py test.

I'll update the commit msg.

e-carlin:
LGTM. Thanks for the info.

PR #722, opened 2024-11-15
707 redhat docker dev

e-carlin:


installers/redhat-docker/radiasoft-download.sh
```diff
@@ -25,23 +25,13 @@ umount /var/lib/docker
 rmdir /var/lib/docker
 perl -pi -e "s{^/var/lib/docker.*}{}" /etc/fstab
 
-Then re-run this command
+Then re-run this command. This should only happen in development environments.
```
robnagler, line 28: Maybe this command should be called redhat-docker-dev? The comment on line 33 is a dev-only comment. This can't be run with rsconf.

installers/redhat-docker/radiasoft-download.sh
```diff
@@ -25,23 +25,13 @@ umount /var/lib/docker
...
 '
     fi
     if selinuxenabled; then
         perl -pi -e 's{(?<=^SELINUX=).*}{disabled}' /etc/selinux/config
         install_err 'Disabled selinux. You need to "vagrant reload", then re-run this installer'
     fi
-    declare o=rhel
-    if install_os_is_fedora; then
```
robnagler, line 36: Is this no longer valid for 36? What adds the repo?

e-carlin: See line 45. This code was repeated code.

installers/redhat-docker/radiasoft-download.sh
e-carlin: Hmm why can't this be run with rsconf? That is exactly what I'm trying to enable. We currently have this code roughly copied in rsconf https://github.com/radiasoft/rsconf/pull/522/files . So, I'm trying to eliminate that duplication.

e-carlin: Ok. The parameters bug me. I'm going to see if I can split out the installer into setup and install. That way rsconf can just call the install part and manage the rest of the setup (ex selinux) itself. 

If that doesn't work I'll use params.

robnagler (approved):

PR #729, opened 2025-01-22
fix #728 add pyDAGMC to openmc python installs

### LINKED ISSUE #728: add PyDAGMC to openmc python installs
https://github.com/svalinn/pydagmc

moellep:


e-carlin (approved):
lgtm. I'll build it in so it goes out with the next alpha.

PR #736, opened 2025-01-31
Fix #735: Don't configure eth1 on almalinux

### LINKED ISSUE #735: vagrant vm trouble connecting over ssh
Deploying vm_devbox keeps failing with a timeout when trying to connect to the vm over ssh during the setup process. It happens at different times each time I run it but it always seems to happen after the vm is rebooted. 

My vm was able to be setup successfully. But, looking at the networking configuration it seems wrong:
```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 08:00:27:e4:4c:af brd ff:ff:ff:ff:ff:ff
    altname enp0s3
    inet <ip1>/24 brd 10.0.2.255 scope global dynamic noprefixroute eth0
       valid_lft 46315sec preferred_lft 46315sec
3: eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 08:00:27:13:c8:0d brd ff:ff:ff:ff:ff:ff
    altname enp0s8
    inet <ip1>/24 brd 10.0.2.255 scope global noprefixroute eth1
       valid_lft forever preferred_lft forever
    inet6 fe80::1c12:9152:924f:e35e/64 scope link noprefixroute
       valid_lft forever preferred_lft forever
```
Both eth0 and eht1 are listening to the same network.

I'm trying to setup another vm_devbox on a new hosts and the install always fails. If I manually access the vm `ip a` reports the same thing. 

We manually configured the network https://github.com/radiasoft/download/blob/897fd299613e466ae9864a421e2d7f95edb66f57/installers/vagrant-dev/radiasoft-download.sh#L65. 

Looking at other vm's of mine (on centos7 host) they don't have two interfaces with the same net. 

Maybe the need to manually configure the net was only needed on centos7? It is inside of the vm so my guess is not. But, on alma we have a newer version of vbox and vagrant so maybe the problem was resolved in one of them. I'm going to limit the manual provision of net to non alma hosts.

e-carlin:
On alma we are running into trouble where two interfaces with the same network were being created. Likely one from vagrant/vbox and one from the nmcli command. This was causing trouble during install of the vm where it could no longer be accessed over ssh.

### DIFF
--- a/installers/vagrant-dev/radiasoft-download.sh
+++ b/installers/vagrant-dev/radiasoft-download.sh
@@ -57,7 +57,7 @@ EOF
 vagrant_dev_eth1() {
     declare os=$1
     declare ip=$2
-    if [[ $os == fedora && $vagrant_dev_private_net ]]; then
+    if ! install_os_is_almalinux && [[ $os == fedora && $vagrant_dev_private_net ]]; then
         # Vagrant doesn't handle NetworkManager correctly so setup eth1
         # https://github.com/hashicorp/vagrant/issues/12762
         cat <<EOF


installers/vagrant-dev/radiasoft-download.sh
```diff
@@ -57,7 +57,7 @@ EOF
 vagrant_dev_eth1() {
     declare os=$1
     declare ip=$2
-    if [[ $os == fedora && $vagrant_dev_private_net ]]; then
```
robnagler, line 60: is `$os == fedora` true for alma?

e-carlin: Most of the time. `$os` is guest os. So for sirepo and workers it is fedora. For rsconf primary it is (or will be) alma. `install_os` is host os. I'll rename `$os` to `$guest_os` to be clearer.

PR #740, opened 2025-02-13
Fix #727: add curl --fail

### LINKED ISSUE #727: curl --fail
Switch to this. Also for the boot scripts.

e-carlin:
Makes it easier to detect errors (e.g., bad GITHUB_TOKEN)

### DIFF
--- a/bin/index.sh
+++ b/bin/index.sh
@@ -20,7 +20,7 @@ index_main() {
             https://api.github.com/repos/radiasoft/download/contents/bin/install.sh
         )
     fi
-    curl --silent --show-error "${a[@]}" | bash ${install_debug:+-x} -s "$@"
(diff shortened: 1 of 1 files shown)


bin/index.sh
```diff
@@ -20,7 +20,7 @@ index_main() {
             https://api.github.com/repos/radiasoft/download/contents/bin/install.sh
         )
     fi
-    curl --silent --show-error "${a[@]}" | bash ${install_debug:+-x} -s "$@"
+    curl --fail --silent --show-error "${a[@]}" | bash ${install_debug:+-x} -s "$@"
```
robnagler, line 23: `set -u` at the top, too.

robnagler (changes requested):

robnagler (approved):

PR #742, opened 2025-02-26
Fix #741: disable memballoon by default when using libvirt

### LINKED ISSUE #741: vm_devbox memballoon=none
memballon is a way for the host to reclaim memory from the guest https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/virtualization_administration_guide/section-libvirt-dom-xml-memory-baloon-device . It is enabled by default with vagrant libvirt. When assigning a large amount of memory this setting seems to cause issues. I have a feeling that ssh is being killed inside of the guest. I could instrument a test by preserving kernel logs between boots, I think. But, doesn't seem worth it to me. We aren't going to have much memory pressure on hosts since vm_devbox is all we'll be running. I'm going to disable. In manual testing this works.

e-carlin:
In high memory situations (>90GB allocated to VM) the VM was crashing. Disabling memballoon fixed this.

### DIFF
--- a/installers/vagrant-dev/radiasoft-download.sh
+++ b/installers/vagrant-dev/radiasoft-download.sh
@@ -198,6 +198,12 @@ EOF
     vagrant_dev_post_install "$update"
 }
 
+vagrant_dev_memballoon() {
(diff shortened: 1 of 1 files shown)


robnagler (commented):
lgtm

PR #816, opened 2025-11-12
fix #815 added httpx module to common

### LINKED ISSUE #815: missing httpx package for stripe
When I removed some old bluesky imports (ab5a2a6), it also removed a sub-dependency of the httpx module which is needed by stripe.

httpx should get added to common.sh

moellep:


robnagler (commented):
lgtm

moellep:
Looks like you have to pip install stripe specially to get httpx included:
```pip install stripe[async]```

https://github.com/stripe/stripe-python/blob/cac5dc933ac31dbec47a02a74c1075f3bd3695af/pyproject.toml#L39

PR #893, opened 2026-06-01
fix #871 TMAP8 

### LINKED ISSUE #871: Add TMAP8 and MOOSE to beamsim
https://mooseframework.inl.gov/TMAP8/

moellep:
This script creates a runnable tmap8 executable. It takes over an hour to build.

The rpm is 295M. Could be reduced by removing symbols?

The build creates sub builds of libmesh and petsc. These don't conflict with the other rpms because they are in ~/.local/libmesh and ~/.local/petsc.  We could try to get the petsc requirements directly into petsc.sh instead?

The TMAP8 repo does not have tags so we should lock down to a specific git commit, rather than the top of main which it is currently using.

A minimal tmap8 test case:

$ tmap8 -i test.i

```
# test.i
diffusivity = 1.0
thickness = 200.0

[Mesh]
  type = GeneratedMesh
  dim = 1
  nx = 10
  xmax = ${thickness}
[]

[Variables]
  [u]
  []
[]

[Kernels]
  [diff]
    type = MatDiffusion
    variable = u
    diffusivity = ${diffusivity}
  []
  [time]
    type = TimeDerivative
    variable = u
  []
[]

[BCs]
  [left]
    type = DirichletBC
    variable = u
    boundary = left
    value = 1
  []
  [right]
    type = DirichletBC
    variable = u
    boundary = right
    value = 0
  []
[]

[Postprocessors]
  [mid_conc]
    type = PointValue
    variable = u
    point = '1 0 0'
    outputs = csv
  []
[]

[Executioner]
  type = Transient
  end_time = 5
  dt = 1
  solve_type = NEWTON
  petsc_options_iname = '-pc_type'
  petsc_options_value = 'lu'
[]

[Outputs]
  file_base = tmap8_test_out
  [csv]
    type = CSV
    execute_on = 'INITIAL TIMESTEP_END'
  []
[]
```

should produce a csv with output:
```
time,mid_conc
0,0
1,0.9369461717326
2,0.9372854335141
3,0.9376203995413
4,0.93795114219532
5,0.93827773245105
```

### DIFF
--- a/installers/rpm-code/codes/tmap8.sh
+++ b/installers/rpm-code/codes/tmap8.sh
@@ -3,7 +3,7 @@
 tmap8_main() {
     codes_yum_dependencies bison flex libtirpc-devel
     codes_dependencies common
-    codes_download idaholab/TMAP8 main
+    codes_download idaholab/TMAP8 ec0413009094eb9efc8c06e5133cd3f63621e051
     declare moose_dir=$PWD/moose
     declare petsc_prefix=${codes_dir[prefix]}/petsc
     declare libmesh_dir=${codes_dir[prefix]}/libmesh
@@ -27,6 +27,7 @@ tmap8_main() {
         -not -path "*/libmesh/*" \
         -name "*.so*" \
         -exec cp --no-dereference --preserve=links {} "$lib_dir/" \;
+    find "$lib_dir" -name "*.so*" ! -type l -exec strip --strip-unneeded {} \;
     # MOOSE and its modules look for data at <prefix>/share/<name>/data at runtime
     install -d -m 755 "${codes_dir[share]}/moose"
     cp -a "$moose_dir/framework/data" "${codes_dir[share]}/moose/"
@@ -36,6 +37,7 @@ tmap8_main() {
         install -d -m 755 "${codes_dir[share]}/$mod"
         cp -a "$d" "${codes_dir[share]}/$mod/"
     done
+    strip --strip-unneeded tmap8-opt
     install -m 555 tmap8-opt "${codes_dir[bin]}/tmap8-bin"
     # Wrapper so the binary finds its libs regardless of the build tree
     declare lp="${lib_dir}:${codes_dir[prefix]}/libmesh/lib:${codes_dir[prefix]}/petsc/lib"
--- a/installers/rpm-code/codes/tmap8.sh
+++ b/installers/rpm-code/codes/tmap8.sh
@@ -1,148 +1,57 @@
 #!/bin/bash
 
-tmap8_libmesh() {
-    declare moose_dir=$1
-    declare petsc_prefix=$2
-    declare libmesh_dir=$3
-    MOOSE_DIR=$moose_dir \
-    PETSC_DIR=$petsc_prefix \
-    LIBMESH_DIR=$libmesh_dir \
-    METHODS=opt \
-    MOOSE_JOBS=$(codes_num_cores) \
-        "$moose_dir"/scripts/update_and_rebuild_libmesh.sh \
-        --skip-submodule-update \
-        --disable-shared \
-        --enable-static \
-        --disable-netgen
-    # libmesh installs bin/ utility programs (meshtool, splitter, compare, etc.),
-    # examples/, contrib/, share/, and etc/ that TMAP8 never uses; with static
-    # PETSc/libmesh each utility separately embeds the full dependency tree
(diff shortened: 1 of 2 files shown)


moellep:
I made an attempt to remove the sub build for petsc but also ran into problems getting our petsc.sh to build in a non-conflicting way with the slepc.sh rpm. 

Trying to separate out slepc from petsc caused a warning: 
```
                                     ***** WARNING *****
  Compiling HPDDM without SLEPc, PCHPDDM won't be available, unless reconfiguring with
  --download-slepc or configuring SLEPc with --download-hpddm
```

So for now I'm keeping the tmap8 petsc subbuild.

moellep:
This is ready for review. tmap8 is statically compiled. Had problems with petsc libceed building statically (its own bundled Makefile only ever builds libceed.so, never a .a), so it remains the only .so needed by tmap8. Several .so files are created for moose's python pyhit library.

installers/rpm-code/codes/tmap8.sh
```diff
@@ -0,0 +1,145 @@
...
+    find "$moose_dir/modules" -maxdepth 2 -name data -type d | while read -r d; do
+        declare mod
+        mod=$(basename "$(dirname "$d")")
+        install -d -m 755 "${codes_dir[share]}/$mod"
+        cp -a "$d" "${codes_dir[share]}/$mod/"
+    done
+    strip --strip-unneeded tmap8-opt
+    install -m 555 tmap8-opt "${codes_dir[bin]}/tmap8-bin"
+    # Wrapper so the binary finds its libs regardless of the build tree
+    declare lp="${lib_dir}:${codes_dir[prefix]}/petsc/lib"
```
robnagler: `lp=$lib_dir:${codes_dir[prefix]}/petsc/lib`

quotes unnecessary as are braces. the hardwire of petsc/lib is problematic. there's a prefix_dir above. I don't understand why this isn't in tmap8/petsc, since it's petsc for tmap8 and nobody else.

A lot of this code is sloppy bash. There's for example install_file_from_stdin which should be used.

installers/rpm-code/codes/tmap8.sh
```diff
@@ -0,0 +1,145 @@
...
+export LD_LIBRARY_PATH="${lp}\${LD_LIBRARY_PATH:+:}\${LD_LIBRARY_PATH:-}"
+exec "${codes_dir[bin]}/tmap8-bin" "\$@"
+EOF
+}
+
+tmap8_moose_python() {
+    declare moose_dir=$1
+    declare hit_src=$moose_dir/framework/contrib/hit
+    declare python_inc
+    python_inc=$(python3 -c 'import sysconfig; print(sysconfig.get_config_var("INCLUDEPY"))')
```
robnagler: codes_python_include_dir

installers/rpm-code/codes/tmap8.sh
```diff
@@ -0,0 +1,145 @@
...
+    # build hit.so (Cython binding to the WASP HIT parser) needed by pyhit
+    CPATH=$python_inc \
+        make -C "$hit_src" -j"$(codes_num_cores)" bindings
+    # rewrite build-tree RUNPATH to the installed WASP lib dir so pyhit
+    # works without LD_LIBRARY_PATH after the RPM is deployed
+    patchelf --set-rpath "${codes_dir[lib]}/tmap8" "$hit_src/hit.so"
```
robnagler: Shouldn't it just use ld_library_path? rpath is problematic (not relocatable)

installers/rpm-code/codes/tmap8.sh
```diff
@@ -0,0 +1,145 @@
...
+    declare site
+    site=$(codes_python_lib_dir)
+    for pkg in pyhit moosetree mooseutils mms; do
+        cp -r "$moose_dir/python/$pkg" "$site/"
```
robnagler: poor coding.

installers/rpm-code/codes/tmap8.sh
```diff
@@ -0,0 +1,145 @@
...
+    done
+    # hit.so must be importable as a top-level module (pyhit does `import hit`)
+    install -m 755 "$hit_src/hit.so" "$site/"
```
robnagler: 555

robnagler (dismissed):
Not really sure what to say here. It's sloppy but it works. I don't want to spend time on this.

PR #894, opened 2026-07-16
perl-dev shouldn't require a private repo

schellj:


installers/perl-dev/radiasoft-download.sh
```diff
@@ -62,12 +62,6 @@ EOF
     fi
     install_source_bashrc
     _bivio_home_env_update -f
-    # Needed for bashrc_b_env_aliases to contain complete set
-    cd ~/src/biviosoftware
-    gcl perl-Artisans
-    ln -s ../biviosoftware/perl-Artisans ~/src/perl/Artisans
-    cd
-    install_source_bashrc
```
robnagler, line 70: you need this line: install_source_bashrc

robnagler (changes requested):

installers/perl-dev/radiasoft-download.sh
schellj: ok, I can put it back. I wasn't sure one would need it because of https://github.com/biviosoftware/home-env/blob/master/install.sh#L158, but that assumes there's no ~/btest-mail and I suppose it doesn't hurt to do the `install_source_bashrc` again.

robnagler: It needs to be there, because the rest of the install does things to those files. It should be done at the very end.

robnagler: And, inside install.sh it's another shell.

robnagler (approved):

PR #897, opened 2026-07-23
Fix #896 replace `install_version_centos` with more accurate name `install_version_rhel`

### LINKED ISSUE #896: `install_version_centos` should be `install_version_rhel`
`install_version_centos` is being used both for the CentOS version and AlmaLinux version. Technically, this is a RHEL version, since the point of Alma is to be mirror of RHEL versions. Let's use RHEL as the variable name going forward.

schellj:
Presumably, existing `/etc/profile.d/rs-redhat-base.sh` files would need to get updated.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -314,7 +314,7 @@ install_init_vars_versions() {
         install_err 'unable to determine operating system version'
     fi
     : ${install_version_fedora:=43}
-    : ${install_version_centos:=7}
+    : ${install_version_rhel:=7}
     if [[ ! ${install_version_python:-} ]]; then
         if install_os_is_centos_7; then
             install_version_python=3.9.15
@@ -629,7 +629,7 @@ install_vars_export() {
         install_proprietary_key
         install_version_fedora
         install_version_python
-        install_version_centos
+        install_version_rhel
         $(compgen -A variable RADIA_RUN_)
         $(compgen -A variable GITHUB_)
     )
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -314,8 +314,7 @@ install_init_vars_versions() {
         install_err 'unable to determine operating system version'
     fi
     : ${install_version_fedora:=43}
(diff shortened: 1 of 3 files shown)


bin/install.sh
```diff
@@ -325,6 +325,15 @@ install_init_vars_versions() {
...
 }
 
+install_init_vars_rhel() {
+    if ! [[ ${install_version_rhel:-} || ${install_version_centos:-} ]]; then
+        install_version_rhel=7
+    elif [[ ${install_version_centos:-} ]]; then
+        # Backward compatibility
+        install_version_rhel=$install_version_centos
+    fi
+}
```
robnagler, line 337: For backwards compat, install_version_centos must be set to what rhel is.

installers/redhat-base/radiasoft-download.sh
```diff
@@ -97,10 +97,10 @@ $(declare -f install_file_from_stdin)
 cat <<'END_CAT' | install_file_from_stdin 444 root root /etc/profile.d/rs-redhat-base.sh
 : \${RADIA_RUN_SERVER:='$install_server'}
 : \${install_depot_server:='$install_depot_server'}
-: \${install_version_centos:='$install_version_centos'}
```
robnagler, line 100: Both need to be set

robnagler (changes requested):

bin/install.sh
schellj: Sorry, I was confused by a previous comment.

Should `install_version_centos` get exported here? https://github.com/radiasoft/download/blob/896-install-version-rhel/bin/install.sh#L643

robnagler: Yes, and btw that link didn't work.

schellj: Oh, it's because the branch is in my fork. Anyway, code fixed.

robnagler (approved):
LGTM

